PR #431, opened 2014-11-13
Fix action bar for reports tool - refs CT#7382

danbarretodev:


jmontoyaa:
This change should be proposed in the 1.10.x branch.
This code is not using the latest code of 1.9.10

PR #451, opened 2014-12-16
Update trad4all.inc.php

forelo:
Added registration mail variable


ywarnier:
I have made the change manually in translate.chamilo.org. Thanks for the PR

PR #464, opened 2015-02-02
Fix bug, audio was sent to root folder, text2audio tool - refs CT#7483

danbarretodev:


main/document/create_audio.php
```diff
@@ -345,7 +345,7 @@ function update_voices(selectedvoicegroup){
  */
 function downloadMP3_google($filepath, $dir)
 {
-	$location='create_audio.php?'.api_get_cidreq().'&id='.Security::remove_XSS($_POST['document_id']).'&dt2a=google';
+    $location='create_audio.php?'.api_get_cidreq().'&id='.Security::remove_XSS($_POST['id']).'&dt2a=google';
```
ywarnier: When you are sure a variable is an integer, please *always* use "intval()" as an integer never generates a risk of XSS.

main/document/create_audio.php
```diff
@@ -420,7 +420,7 @@ function downloadMP3_google($filepath, $dir)
  * @version january 2011, chamilo 1.8.8
  */
 function downloadMP3_pediaphon($filepath, $dir){
-	$location='create_audio.php?'.api_get_cidreq().'&id='.Security::remove_XSS($_POST['document_id']).'&dt2a=pediaphon';
+	$location='create_audio.php?'.api_get_cidreq().'&id='.Security::remove_XSS($_POST['id']).'&dt2a=pediaphon';
```
ywarnier: Same as above. Use intval() instead of remove_XSS, when possible

PR #467, opened 2015-02-03
Remove duplicated Qti classes and methods - refs #7504

AngelFQC:


jmontoyaa:
The problem stills appears. For example the class "ImsAnswerMultipleChoice" exists inside 
qti_classes.php and qti2_classes.php

AngelFQC:
ImsAnswerMultipleChoice class in qti2_classes.php will be used?

jmontoyaa:
That's the main issue of the task. Identify why we have 2 classes like this.

PR #468, opened 2015-02-04
Add extra content in admin panels - refs BT#9325

AngelFQC:


main/admin/index.php
```diff
@@ -26,6 +26,21 @@
 
 $nameTools = get_lang('PlatformAdmin');
 
+if (api_is_multiple_url_enabled()) {
+    $accessUrlId = api_get_current_access_url_id();
+
+    if ($accessUrlId != -1) {
+        $urlInfo = api_get_access_url($accessUrlId);
+        $url = api_remove_trailing_slash(preg_replace('/https?:\/\//i', '', $urlInfo['url']));
+        $cleanUrl = replace_dangerous_char($url);
```
ywarnier: I believe the URL is safe anyway (it comes from the access_url table, and all entries to that table are filtered), so no need to replace_dangerous_char() here.

main/inc/ajax/admin.ajax.php
```diff
@@ -35,6 +35,74 @@
...
+        if (api_is_multiple_url_enabled()) {
+            $accessUrlId = api_get_current_access_url_id();
+
+            if ($accessUrlId == -1) {
+                die;
+            }
+
+            $urlInfo = api_get_access_url($accessUrlId);
+            $url = api_remove_trailing_slash(preg_replace('/https?:\/\//i', '', $urlInfo['url']));
+            $cleanUrl = replace_dangerous_char($url);
```
ywarnier: Same here. I don't think we need to replace_dangerous_char()

main/inc/ajax/admin.ajax.php
```diff
@@ -35,6 +35,74 @@
...
+            $cleanUrl = str_replace('/', '-', $cleanUrl);
+
+            $newUrlDir = api_get_path(SYS_PATH) . "home/$cleanUrl/admin/";
+        } else {
+            $newUrlDir = api_get_path(SYS_PATH) . "home/admin/";
+        }
+
+        if (!is_dir($newUrlDir)) {
+            mkdir($newUrlDir, api_get_permissions_for_new_directories(), true);
```
ywarnier: If making a system call, you should always choose between two possibilities: 1) recover the return value in a variable and suffocate the error (with "@"), or 2) manage exceptions. Either way, the current mkdir(), unlink(), touch() and file_put_contents() calls are all dangerous as they could fail and show too much information to potential attackers.

main/inc/ajax/admin.ajax.php
```diff
@@ -35,6 +35,74 @@
     case 'version':
         echo version_check();
         break;
+
+    case 'save_block_extra':
+        $content = isset($_POST['content']) ? Security::remove_XSS($_POST['content']) : null;
+        $blockId = isset($_POST['block']) ? Security::remove_XSS($_POST['block']) : null;
```
ywarnier: Watch out! $blockId is a string, so it would be better to call it "$blockName" or something like that.
Because you use $blockId to form part of a path to a file, it could be used as a path traversal vulnerability (for example to get to /etc/passwd). Make sure you use Security::check_rel_path() to compare the final path to api_get_path(SYS_PATH).'home/'

AngelFQC:
PR actualizado

PR #469, opened 2015-02-04
Bt9068a

ilosada:


main/css/base.css
```diff
@@ -5304,3 +5304,146 @@ i.size-32.icon-new-work{
 {
     height: 52px;
```
jmontoyaa: To ignore permissions changes as in file "archive/.htaccess"

You should use:

git config core.fileMode false

http://stackoverflow.com/questions/1580596/how-do-i-make-git-ignore-file-mode-chmod-changes

ywarnier: Just told him that :-) Done.

main/survey/create_new_survey.php
```diff
@@ -147,6 +147,13 @@ function setFocus(){
 $form->addElement('date_picker', 'end_date', get_lang('EndDate'));
 
 $form->addElement('checkbox', 'anonymous', null, get_lang('Anonymous'));
+$visibleResults = array(
+    SURVEY_VISIBLE_TUTOR => get_lang('Coach'),
+    SURVEY_VISIBLE_TUTOR_STUDENT => get_lang('Coach').' '.get_lang('And').' '.get_lang('Student'),
```
jmontoyaa: There should be a new variable "CoachAndStudent".

main/survey/survey.lib.php
```diff
@@ -5189,4 +5323,62 @@ static function show_link_available($user_id, $survey_code, $user_answer)
...
         }
     }
+
+    /**
+     * Display survey question chart
+     * @param	array	Chart data
+     * @param	boolean	Tells if the chart has a serie. False by default
+     * @return	void 	(direct output)
+     */
+    static function drawChart($chartData, $hasSerie = false)
```
jmontoyaa: Function visibility  missing "public".

main/survey/survey.lib.php
```diff
@@ -5189,4 +5323,62 @@ static function show_link_available($user_id, $survey_code, $user_answer)
...
+    {
+        $htmlChart = '<div id="chartContainer" class="span12">'.
```
jmontoyaa: This code smells... Is not better to use a template?

main/survey/survey.lib.php
```diff
@@ -5189,4 +5323,62 @@ static function show_link_available($user_id, $survey_code, $user_answer)
...
+            api_get_js("d3/d3.v3.4.8.js").'
+            <script src="http://dimplejs.org/dist/dimple.v2.1.0.min.js"></script>
```
jmontoyaa: Not a good idea to use external js like this. 
The lib should be added in Chamilo if it has a GPLv3 compatible license.

main/survey/fillsurvey.php
```diff
@@ -330,7 +338,7 @@
     $paged_questions_sec = array();
 
     if (!empty($survey_data['survey_introduction'])) {
-        echo '<div id="survey_content" class="survey_content">'.$survey_data['survey_introduction'].'</div>';
+        echo '<div id="survey_content" class="survey_content">'.strip_tags($survey_data['survey_introduction']).'</div>';
```
jmontoyaa: Not sure about this... what happen if the introduction had tags (via ckeditor)?

main/survey/survey.lib.php
ywarnier: // Copyright: 2014 PMSI-AlignAlytics
// License: "https://github.com/PMSI-AlignAlytics/dimple/blob/master/MIT-LICENSE.txt"
// Source: /src/objects/begin.js

MIT license is OK for inclusion in Chamilo. Please download and put into main/inc/lib/javascript/

main/survey/survey.lib.php
ywarnier: Also... SVG is not supported by IE8 (which we still support with Chamilo). Isn't this Dimple thing going to generate SVG-only?
I agree that the code looks a bit too "generated by hand" to feel safe.

main/survey/survey.lib.php
```diff
@@ -3008,31 +3049,43 @@ public static function display_question_report($survey_data)
...
+		            }
+
+		            echo '      </td>';
+                    echo '		<td class="center">'.round($answers_number, 2).' %</td>';
+                    echo '		<td class="center">';
                     $size = $answers_number*2;
                     if ($size > 0) {
                         echo '<div style="border:1px solid #264269; background-color:#aecaf4; height:10px; width:'.$size.'px">&nbsp;</div>';
+                    }else{
+                        echo '<div style="text-align: left;">'.get_lang("NoDataRegistered").'</div>';
```
ywarnier: Please use the existing "NoDataAvailable"

main/survey/reporting.php
```diff
@@ -17,6 +17,7 @@
 require_once '../inc/global.inc.php';
 require_once 'survey.lib.php';
 $this_section = SECTION_COURSES;
+$cidReq = Security::remove_XSS(api_get_cidreq());
```
ywarnier: Because of where cidReq passes through before and also during api_get_cidreq(), you don't need to remove_XSS() on the result of api_get_cidreq().

main/survey/survey.lib.php
```diff
@@ -2877,6 +2897,7 @@ function jumpMenu(targ,selObj,restore) {
      * @todo the pagebreak and comment question types should not be shown => removed from $survey_data before
      * @author Patrick Cool <patrick.cool@UGent.be>, Ghent University
      * @version February 2007 - Updated March 2008
+     * @CSS Alex Aragon <alex.aragon@beeznest.com> - update January 2015
```
ywarnier: This is not a recognized phpdoc tag. Additionally, a function should only have an @author tag if it is very important to be able to identify the author (for example, it is a very complex function which depends on stuff that we are not sure about). Otherwise, no @author tag, and certainly not just for CSS changes (which should be located outside of a php script anyway)

main/survey/survey.lib.php
```diff
@@ -2902,17 +2918,27 @@ public static function display_question_report($survey_data)
...
-                } else {
-                    echo $i;
-                }
-                if ($i < $survey_data['number_of_questions']) {
-                    echo ' | ';
+            echo '<div id="question_report_questionnumbers" class="pagination">';
+            /* echo '<ul><li class="disabled"><a href="#">'.get_lang('Question').'</a></li>'; */
+
+            if ($currentQuestion != 0 ) {
+                echo '<li><a href="'.api_get_path(WEB_CODE_PATH).'survey/reporting.php?action='.Security::remove_XSS($_GET['action']).'&amp;'.api_get_cidreq().'&amp;survey_id='.Security::remove_XSS($_GET['survey_id']).'&amp;question='.Security::remove_XSS($offset-1).'">'.get_lang('PreviousQuestion').'</a></li>';
```
ywarnier: If we know for sure that a variable is an integer, then the intval() function should be called on that variable instead of any more complex form of Security::remove*()

main/survey/create_new_survey.php
ywarnier: I added the CoachAndStudent variable, so this just needs to change here

PR #472, opened 2015-02-04
Allow assign description to session - refs BT#9049

AngelFQC:


plugin/advancedsessions/lang/english.php
```diff
@@ -0,0 +1,3 @@
+<?php
+$strings['plugin_title'] = "Advanced Sessions";
+$strings['plugin_comment'] = "Add a session extra field for set a session description";
```
ywarnier: for set -> to set

plugin/advancedsessions/src/AdvancedSessionsPlugin.class.php
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * The Advanced Session allow add sessions' extra fields 
+ * @author Angel Fernando Quiroz Campos <angel.quiroz@beeznest.com>
+ * @package chamilo.plugin.advancedSessions
+ */
+class AdvancedSessionsPlugin extends Plugin
+{
+
+    const FIELD_NAME = 'as_description';
```
ywarnier: "as" should be more expressive, like "advsess" or something like that, because as is an English word

ywarnier:
Generally good-looking code.

main/admin/resume_session.php
```diff
@@ -10,6 +10,12 @@
 $cidReset = true;
 require_once '../inc/global.inc.php';
 
+$advancedSessionsPluginFilePath = api_get_path(PLUGIN_PATH) . 'advancedsessions/src/AdvancedSessionsPlugin.class.php';
```
jmontoyaa: You don't need to do this. Just run an composer update and it will add the class in the autoload.

jmontoyaa: The same comment for the rest of files.

main/session/index.php
```diff
@@ -354,6 +360,22 @@
 echo Display::tag('h1', $session_info['name']);
 echo $dates.'<br />';
 
+if (class_exists('AdvancedSessionsPlugin') && AdvancedSessionsPlugin::hasDescriptionField()) {
```
jmontoyaa: Don't need also to add this "class_exists('AdvancedSessionsPlugin')"

AngelFQC:
PR actualizado

main/template/default/auth/sessions_catalog.tpl
```diff
@@ -87,65 +87,70 @@
...
-                                             <a class="accordion-toggle" data-toggle="collapse" data-parent="#session-{{ session.id }}-accordion" href="#session-{{ session.id }}-courses">
-                                                {{ 'CourseList' | get_lang }}
-                                            </a>
-                                        </div>
-                                        <div id="session-{{ session.id }}-courses" class="accordion-body collapse in">
-                                            <div class="accordion-inner"></div>
-                                        </div>
+                            <div class="span10 border-info">
+                                <h3>{{ session.name }}</h3>
+                                <div class="tutor"><img src="{{ _p.web }}/main/img/teachers.gif" width="16px"> {{ 'GeneralCoach' | get_lang }} {{ session.coach_name }}</div>
```
ywarnier: Avoid .gif icons. Prefer .png versions in main/img/icons/. Maybe there is a returnIcpnPath modifier? There should be one, in any case.

main/admin/resume_session.php
```diff
@@ -244,6 +244,20 @@
 
 }
 ?>
+<?php
+if (class_exists('AdvancedSessionsPlugin') && AdvancedSessionsPlugin::hasDescriptionField()) {
```
ywarnier: Would be better to check if the plugin is enabled (in terms of efficiency).

ywarnier:
I asked Angel to move this to core. Given the only effect it has is to add a description field to the session table, I cannot see why it would need to be a plugin. Initially, we thought we wouldn't be able to change the database, which is why we started it as a plugin

AngelFQC:
#480

PR #473, opened 2015-02-04
Restriction profile editing - refs BT#9062

AngelFQC:


ywarnier:
No. Un DRH = Director de Recursos Humanos. A session admin = Session admin. The task (9062) says the *DRH* cannot edit. Nothing is said about the session admin.

AngelFQC:
El DRH no tiene acceso a la edición de perfil. El sessión admin sí
Respondí en 9062

ywarnier:
Rejected. Does not really implement what's expected in BT#9062

PR #476, opened 2015-02-05
Bt9083

ilosada:


jmontoyaa:
Please provide a title in the PR instead of just the code Bt9083.

ywarnier:
Not generic enough. Too Drupal-specific. This should use hooks, really, and a separate Chamilo plugin to use this hook. I've merged the hook system from Daniel, so please use that.

PR #477, opened 2015-02-05
Bt9194

ilosada:


ywarnier:
This seems to include PR476, so same observations here. Too specific + use hook system to create a plugin (drupal_ws_user_sync for example) that we can include and configure apart from the chamilo core (and, I insist, that uses the hook system).

PR #478, opened 2015-02-05
Display courses without session in report - refs BT#9087

AngelFQC:


main/inc/lib/usermanager.lib.php
```diff
@@ -5005,4 +5005,47 @@ public static function getUsersByOfficialCode($officialCode)
...
+            'AND session_id = ? ' => $sessionId
+        );
+
+        if (!empty($from) && !empty($until)) {
+            $whereConditions["AND (login_course_date >= '?' "] = $from;
+            $whereConditions["AND logout_course_date <= DATE_ADD('?', INTERVAL 1 DAY)) "] = $until;
+        }
+
+        $trackResult = Database::select(
+            'SEC_TO_TIME(SUM(UNIX_TIMESTAMP(logout_course_date) - UNIX_TIMESTAMP(login_course_date))) as total_time',
```
jmontoyaa: I think that SEC_TO_TIME is mysql specific. It should be better to do it in PHP.

There's a function that does something like this in  main/newscorm  "calculateScormTime()"
Maybe it should be a good idea to create a function that does this in main_api.lib

ywarnier: Agreed. If it does the same, then move the function out of main/newscorm and into main_api.lib

AngelFQC:
PR actualizado

PR #492, opened 2015-02-13
Kannel SMS

ilosada:


ywarnier:
There are many things we will be able to improve on the basis of this PR (like move to using hooks and allow several plugins to interact at this level) but this is a first good step forward.

PR #496, opened 2015-02-13
Add field to check whether generate certificates - refs #7525

AngelFQC:


main/gradebook/lib/be/category.class.php
```diff
@@ -350,6 +369,7 @@ private static function create_root_category()
         $cat->set_parent_id(null);
         $cat->set_weight(0);
         $cat->set_visible(1);
+        $cat->setGenerateCertificates(false);
```
jmontoyaa: This should be:

$cat->setGenerateCertificates(0);

instead of

$cat->setGenerateCertificates(false);

According the the PHPdoc

PR #499, opened 2015-02-16
Fix 'buy courses' plugin bugs and glitches

ilosada:


plugin/buycourses/src/function.php
```diff
@@ -230,54 +222,40 @@
...
-    if ($priceMin != '') {
-        if ($filter == '') {
-            $filter .= "a.price >= '" . $priceMin . "'";
-        } else {
-            $filter .= " AND a.price >= '" . $priceMin . "'";
-        }
+    $sql = "SELECT a.course_id, a.visible, a.price, b.*
+            FROM $tableBuyCourse a, $tableCourse b
+            WHERE a.course_id = b.id AND a.session_id = 0
+            AND a.visible = 1;";
```
ywarnier: No need to include ";" inside the query string. Just saying. No need to fix, but in general, it is unnecessary.

plugin/buycourses/src/function.php
```diff
@@ -230,54 +222,40 @@
...
-        $sql = "SELECT a.course_id, a.visible, a.price, b.*
-            FROM $tableBuyCourse a, $tableCourse b
-            WHERE a.course_id = b.id AND a.session_id = 0
-            AND a.visible = 1 AND " . $filter . ";";
+    /*if (!empty($category)) {
+        $filter .= " AND b.category_code = '".$category."'";
+    }*/
+
+    if (!empty($filter)) {
+        $sql = substr_replace($sql, $filter.";", -1);
```
ywarnier: Same here

plugin/buycourses/src/function.php
```diff
@@ -288,7 +266,7 @@
             FROM $tableCourseRelUser a, $tableUser b
             WHERE a.course_code = '" . $row['code'] . "'
             AND a.role <> ''
-            AND a.role <> 'NULL'
+            AND a.role IS NOT NULL
             AND a.user_id = b.user_id;";
```
ywarnier: And here

plugin/buycourses/src/function.php
```diff
@@ -330,41 +308,52 @@
...
+                            href="'.$server.'plugin/buycourses/src/ajax.php?
+                            a=show_course_information&code='.$course['code'].'">
+                                <img alt="" src="'.$server.$course['course_img'].'">
+                            </a>
+                        </div>
+                    </div>
+                    <div class="span3">
+                        <div class="categories-course-description">
+                            <h3>'.$course['title'].'</h3>
+                            <h5>'.get_lang('Teacher').': '.$course['teacher'].'</h5>
```
ywarnier: Try to use a single variable with "Teacher: %s" for example

PR #502, opened 2015-02-16
Categorías de sesiones [x2.1.3.6]a

ilosada:


tests/scripts/insert_sessions_categories.php
```diff
@@ -0,0 +1,16 @@
+<?php
```
jmontoyaa: License

tests/scripts/insert_sessions_categories.php
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * This script generates four session categories.
+ */
+require_once '../../main/inc/global.inc.php';
+api_protect_admin_script();
+$categories = array(
```
jmontoyaa: Don't use Spanish

ywarnier: This is a short example script (we needed these categories names specifically). Given it is in tests/scripts/ I don't think the rules with content name really need to apply. The rest still does (license, security, formatting, variables names, etc). Just that the content is specific.

jmontoyaa: MMMmmm ok. Good to know!

PR #504, opened 2015-02-17
Fix sql syntax error when listing users of a session

emarguin:
Hi, here is a fix for a sql error syntax in function get_users_by_session


jmontoyaa:
:+1:

PR #509, opened 2015-02-18
re-activate ldap user search feature using new ldap libs

emarguin:
Here is a pull request to re-activate the old admin's feature allowing to search a user from ldap directory and import him in Chamilo.
I guess it would have been a better way to rewrite the full feature since the code is quite old but I miss time for that.
Anyway : it works fine now and every php notices has been removed too.


ywarnier:
Totally agree with this one. Furthermore, I was in the middle of LDAP stuff myself and was doing part of the same. Thanks.

PR #511, opened 2015-02-18
BT#9413 - Add right click modal to profile skill wheel

ilosada:


ywarnier:
I'm putting this one on hold until the advanced subscription plugin is included.
Also, the css/forge/ style should *not* be included in the PR. The template/default/ changes either.

PR #512, opened 2015-02-18
Create web service to search sessions - refs BT#9418

AngelFQC:


main/inc/lib/sessionmanager.lib.php
```diff
@@ -5430,4 +5430,66 @@ public static function getCoursesListByCourseCoach($coachId)
...
+                "OR id IN (
+                    SELECT session_id
+                    FROM $sfvTable
+                    WHERE field_value LIKE %?%
+                ) " => $term
+            )
+        ));
+
+        $fieldsArray = array(
+            'short_description', 'mode', 'human_text_duration', 'vacancies', 'brochure', 'target', 'schedule'
```
ywarnier: You cannot set fixed extra fields.
I would suggest you add one parameter to the function ($alsoSearchExtraFields, false by default) to decide whether you should search in extra fields or not *and* that you search *all* session extra fields.

main/inc/lib/sessionmanager.lib.php
```diff
@@ -5430,4 +5430,87 @@ public static function getCoursesListByCourseCoach($coachId)
         ));
     }
 
+    /**
+     * Get a session list filtered by name, description or short description extra field
```
ywarnier: This should be updated now, given that "short description" is just a parameter in the array

main/inc/lib/sessionmanager.lib.php
Morpheus2015: Yannick, call me inbox?
Limpeza.Genial@gmail.com

Enviado via IPhone de Limpeza Genial
Em 18/02/2015 19:42, "Yannick Warnier" <notifications@github.com> escreveu:

PR #514, opened 2015-02-18
7514 - Set autofocus on first row score

ilosada:


ywarnier:
Excellent. A huge time saver.

PR #516, opened 2015-02-20
Add conversion from single answer to multiple answer and viceversa - refs #7485

ilosada:


ywarnier:
I have merged only the commit that provides the feature because the rest had apparently already been fixed. Thanks.

PR #520, opened 2015-02-23
B#9323 - Add 'DISTINCT' to session list query to avoid returning repeated records (1.10.x)

ilosada:


ywarnier:
I only commited the change that fixes the query.

PR #522, opened 2015-02-23
B#9324 - Show only session admin created users in user list and in session creation (1.10.x)

ilosada:


ywarnier:
Only included the right commit and added database queries manually

PR #524, opened 2015-02-24
Update advsub plugin

danbarretodev:
Update Hook php doc
Add cron to notify by mail and required scripts


main/inc/lib/sessionmanager.lib.php
```diff
@@ -5794,4 +5806,18 @@ public static function searchSession($term, $extraFieldsToInclude = array())
 
         return $resultData;
     }
+
```
ywarnier: Missing documentation (phpdoc) here

plugin/advanced_subscription/src/AdvancedSubscriptionPlugin.php
```diff
@@ -783,7 +883,7 @@ public function getQueueUrl($params)
             'e=' . intval($params['newStatus']) . '&' .
             'u=' . intval($params['studentUserId']) . '&' .
             'q=' . intval($params['queueId']) . '&' .
-            'is_connected=' . intval($params['is_connected']) . '&' .
+            'is_connected=' . true . '&' .
```
ywarnier: I don't think this will work... what's the string value of the boolean true?

plugin/advanced_subscription/src/scripts/insert_session_fields.php
```diff
@@ -0,0 +1,203 @@
...
+    'field_type' => ExtraField::FIELD_TYPE_TEXT,
+    'field_variable' => 'human_text_duration',
+    'field_display_text' => get_lang('DurationInWords'),
+    'field_visible' => 1,
+    'field_changeable' => 1
+));
+
+$showStatusOptions = array(
+    get_lang('Open'),
+    get_lang('InProcess'),
```
ywarnier: This variable is missing from translate.chamilo.org. I'm adding it now.

plugin/advanced_subscription/src/AdvancedSubscriptionPlugin.php
danbarretodev: tested the generated url and is set to "1" value

main/inc/lib/sessionmanager.lib.php
danbarretodev: Oh, I fogot it, ok, adding phpdoc

PR #527, opened 2015-02-25
Add web service to get a session by its id - refs BT#9465

AngelFQC:


ywarnier:
Try to always send code-style changes and logic changes in separate commits, please.

PR #533, opened 2015-02-26
Update phpdoc of content_to_html method from class PDF - refs BT9461

danbarretodev:


main/inc/lib/pdf.lib.php
```diff
@@ -340,11 +340,14 @@ public function html_to_pdf(
...
-     * @param   string  $css CSS content of a CSS file
-     * @param   string  $pdf_name pdf name
-     * @param   string  $course_code course code
+     * @param   string $document_html valid html
+     * @param   string $css CSS content of a CSS file
+     * @param   string $pdf_name pdf name
+     * @param   string $course_code course code
      * (if you are using html that are located in the document tool you must provide this)
+     * @param string $outputMode the output mode can be:
+     * 'I', 'D', 'F', 'S'
```
jmontoyaa: 'I', 'D', 'F', 'S'
What are the meaning of those? 
We should use complete words.

ywarnier: Unless these are the parameters MPDF expects...
I added documentation for these next to the letters

PR #535, opened 2015-02-26
Certificates report and downloads - refs BT#9442

AngelFQC:


AngelFQC:
Close 1.9.x RP
Open #539 for v1.10.x

PR #536, opened 2015-02-27
Show error message to choose either session duration or time limit - refs BT#9464

AngelFQC:


main/admin/session_edit.php
```diff
@@ -295,8 +295,8 @@
     $description = $params['description'];
     $showDescription = isset($params['show_description']) ? 1: 0;
 
-    $end_limit = $params['end_limit'];
-    $start_limit = $params['start_limit'];
+    $end_limit = isset($params['end_limit']);
+    $start_limit = isset($params['start_limit']);
```
ywarnier: If you change $end_limit and $start_limit to a bool, then you should also check all related uses of these variables afterwards...
The empty($end_limit) doesn't make much sense (false if true, true if false). A !$end_limit is more expressive (and probably faster as well).

PR #538, opened 2015-02-27
Company reports refer only to own users - refs BT#9441

AngelFQC:


main/inc/ajax/model.ajax.php
```diff
@@ -186,7 +187,38 @@ function getWhereClause($col, $oper, $val)
             if (empty($userIdList) || empty($courseCodeList)) {
                 exit;
             }
+        } else if(api_is_student_boss()) {
```
ywarnier: Apart from this missing space... this PR looks perfect!

PR #539, opened 2015-02-27
Certificates report and downloads - refs BT#9442

AngelFQC:


ywarnier:
Apparently there are a series of issues with this PR, probably with other commits which have changed things since then. Please review and re-submit. In particular, main/gradebook/certificate_report.php is saying: Fatal error: Call to a member function get_id() on a non-object in main/gradebook/certificate_report.php on line 78

Also, the certificates list has disappeared from inside the course (main/gradebook/index.php). It has to remain there. The other place is OK to put it, but it has to remain in gradebook/index.php as well.

AngelFQC:
The certificates list still there. Maybe you need to enable certificate generation

PR #541, opened 2015-03-02
Fix save exercise feedback either advanced editor or textarea - refs #7555

AngelFQC:


ywarnier:
I cannot test it at the moment because I cannot take a test: the "save and go to next question" button returns error (I think it is due to a missing exe_id which apparently isn't field as it should in the form.

PR #542, opened 2015-03-02
BT#9437 - create_course_sessions cron script

ilosada:
- Add create_course_sessions cron script
- Add active column to admin query
- Add 'getCoursesWithoutSession' function to course.lib.php


main/cron/create_course_sessions.php
```diff
@@ -0,0 +1,80 @@
...
+function getMonthFirstAndLastDates($initialDate = null) {
+    $startDate = $initialDate ? $initialDate : date("Y-m-01");
+    $nextMonthStartDate = date("Y-m-d", strtotime($startDate." + 1 month"));
+    $endDate = date("Y-m-d", strtotime($nextMonthStartDate." - 1 minute"));
+    return array('startDate' => $startDate, 'endDate' => $endDate);
+}
+
+function createCourseSessions($courses, $administratorId, $startDate, $endDate) {
+    echo "\n";
+    echo $courses ?
```
ywarnier: This condition is too weakly expressed. Do you mean if (count($courses) > 0)) ? Then it should be better expressing it this way!

main/cron/create_course_sessions.php
```diff
@@ -0,0 +1,80 @@
...
+if (php_sapi_name() != 'cli') {
+    exit; //do not run from browser
+}
+
+require_once __DIR__ . "/../inc/global.inc.php";
+
+// First day of the current month to create sessions and add courses for the next month (e.g. "07")
+define("OFFSET", "15");
+
+function getMonthFirstAndLastDates($initialDate = null) {
```
ywarnier: Always give it a little PHPDoc block, please

main/cron/create_course_sessions.php
```diff
@@ -0,0 +1,80 @@
...
+    $startDate = $initialDate ? $initialDate : date("Y-m-01");
+    $nextMonthStartDate = date("Y-m-d", strtotime($startDate." + 1 month"));
+    $endDate = date("Y-m-d", strtotime($nextMonthStartDate." - 1 minute"));
+    return array('startDate' => $startDate, 'endDate' => $endDate);
+}
+
+function createCourseSessions($courses, $administratorId, $startDate, $endDate) {
```
ywarnier: Always give it a little PHPDoc block, please

main/cron/create_course_sessions.php
```diff
@@ -0,0 +1,80 @@
...
+    echo "\n";
+    echo $courses ?
+        "Creating sessions and adding courses for the period between ".$startDate." and ".$endDate :
+        "Every course is already in session for the period between ".$startDate." and ".$endDate;
+    echo "\n=====================================================================================\n\n";
+    // Loop through courses creating one session per each and adding them
+    foreach ($courses as $course) {
+        $sessionName = $course['title']." (".date("M Y", strtotime($startDate)).")";
```
ywarnier: Probably better to use "03/2015" than "03 2015" ;-)
Please use the api_strtotime or api_get_local_time() to avoid issues with moving back to the month before in some cases.

main/cron/create_course_sessions.php
ilosada: I did this way because I see that '$courses' is cleaner and more efficient than '(count($courses) > 0))'.

main/cron/create_course_sessions.php
ilosada: date("M Y") returns the month as verbose, so the result would be "Mar 2015" instead of "03 2015". This line is just to give the new session a friendlier name.

ywarnier: Good point

main/cron/create_course_sessions.php
ywarnier: I did a little check on (bool) count(array()) and (bool) count(array(0 => false)) vs (bool) array() and I'm OK with your change. Good.

PR #543, opened 2015-03-02
Fix the insert admin and anonymous users expiration date on installation

AngelFQC:
Fix 1bcb584


ywarnier:
I want a discussion on this as soon as possible, because I don't get what's causing this and what it could trigger for people upgrading from previous versions of Chamilo

PR #546, opened 2015-03-03
B#9444 - Add remind_course_expiration cron script 

ilosada:


main/cron/remind_course_expiration.php
```diff
@@ -0,0 +1,120 @@
...
+        );
+        foreach ($sessions as $session) {
+            $daysRemaining = date_diff($today, date_create($session['date_end']));
+            $body = sprintf(
+                get_lang('MailCronCourseExpirationReminderBody', null, $platformLanguage),
+                $userCompleteName,
+                $session['name'],
+                $session['date_end'],
+                $daysRemaining->format("%d"),
+                $_configuration['root_web'],
```
ywarnier: I'll have to hint you on how to support multi-url here ($_configuration['root_web'] doesn't).

main/cron/remind_course_expiration.php
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Initialization
+ */
+if (php_sapi_name() != 'cli') {
+    exit; //do not run from browser
+}
+
+// Days before expiration date to send reminders
+define("OFFSET", 2);
+$today = date("Y-m-d");
```
ywarnier: Although date() will use the server's date, this is not necessarily UTC (the server date might be UTC+2, but the database stores times in UTC+0), so you should always use the api_get_utc_datetime() (or at least gmdate()).

PR #554, opened 2015-03-04
B#9022 - Cron script that adds gradebook certificates (1.10.x)

ilosada:


ywarnier:
It looks like changes are missing to registration.soap.php (see comments for 1.9.x PR)

PR #557, opened 2015-03-06
B#9431 - Desarrollar noción de empresa + área

ilosada:
- Replace group name with full hierarchy group text
- Add several group web services


ywarnier:
Just a few comments to review.
In general terms, please avoid sending the "cleanup" changes in the same commit. It makes it more difficult to check.

PR #558, opened 2015-03-06
B#9463 - Apply fixed width to decision column

ilosada:


plugin/advanced_subscription/views/admin_view.tpl
```diff
@@ -40,7 +40,7 @@
                         <th>{{ "InscriptionDate" | get_plugin_lang('AdvancedSubscriptionPlugin') }}</th>
                         <th>{{ "Area" | get_plugin_lang('AdvancedSubscriptionPlugin') }}</th>
                         <th>{{ "BossValidation" | get_plugin_lang('AdvancedSubscriptionPlugin') }}</th>
-                        <th>{{ "Decision" | get_plugin_lang('AdvancedSubscriptionPlugin') }}</th>
+                        <th id="decisionColumn">{{ "Decision" | get_plugin_lang('AdvancedSubscriptionPlugin') }}</th>
```
ywarnier: I would prefer a class than an ID (an ID can only occur once per page, and we might add other tables to this page in the future). CSS classes in general "should" (there's no hard coding convention on that) be named lowercase with "-" to separate words

PR #560, opened 2015-03-09
prerequisite form : check input for scorm too

emarguin:
when a prerequisite is defined by scorm packaged, it's not shown on the prerequisite form due to the fact that Chamilo uses int while scorm use string (ref field). It seems that it's handled everywhere but there.


ywarnier:
I'll rely on you for this one. I hope you tested it enough with int values as well!

emarguin:
well tested but only with a few packages I have. It might be usefull to test a few else, but I don't see any reason why it could break something

PR #561, opened 2015-03-09
Check the completed induction courses - refs BT#9092

AngelFQC:


plugin/advanced_subscription/src/AdvancedSubscriptionPlugin.php
```diff
@@ -208,20 +208,9 @@ public function isAllowedToDoRequest($userId, $params = array(), $collectErrors
...
-        if ($checkInduction && !$completedInduction) {
-            $this->errorMessages[] = $this->get_lang('AdvancedSubscriptionIncompleteInduction');
-
-            if (!$collectErrors) {
-                throw new \Exception($this->get_lang('AdvancedSubscriptionIncompleteInduction'));
-            }
-        }
-
+        $yearlyCostLimit = $plugin->get('yearly_cost_limit');
         $uitMax = $plugin->get('yearly_cost_unit_converter');
```
ywarnier: Please call $uitMax something like $maxCost to avoid being too specific on a language term (uit goes for fiscal tax unit, but it mainly means a cost limit converter here)

main/gradebook/lib/be/category.class.php
```diff
@@ -1636,10 +1636,6 @@ public static function register_user_certificate($category_id, $user_id)
             return false;
         }
 
-        $alleval_course = $category->get_evaluations($user_id, true);
```
ywarnier: I'm not sure about these changes...

AngelFQC: Could not generate certificates having categories; fixed

PR #562, opened 2015-03-10
B#9528 - Add automatic session subscription feature to session catalog

ilosada:


main/auth/courses_controller.php
```diff
@@ -483,14 +483,19 @@ public function getCoursesCategoriesBlock($code = null, $hiddenLinks = false, $l
...
+     * @param string $sessionData The session data
      * @return string The button
      */
-    public function getRegisterInSessionButton($sessionName)
+    public function getRegisterInSessionButton($sessionData)
     {
-        $sessionName = urlencode($sessionName);
+        global $_configuration;
+
+        $sessionData = urlencode($sessionData);
```
ywarnier: That's not enough to make it secure. Please use Security::removeXSS()

PR #590, opened 2015-03-18
C#4507 - Add missing configuration parameters

ilosada:


ywarnier:
Wrong place (should be in main/install/)

PR #646, opened 2015-04-06
Don't show button to export certificates to PDF - refs BT#9237

AngelFQC:


ywarnier:
Instead of removing the student option, please add the new option but leave the previous (student) one. This way, we get the extra flexibility.

AngelFQC:
Updated

PR #649, opened 2015-04-08
C#7233 - Correct glossary word boundary accents issue

ilosada:


ywarnier:
I'll accept this change, but I feel like there must be a better way to do that.

PR #676, opened 2015-04-17
Add skill management on admin index page - refs BT#9583 #7385

AngelFQC:


ywarnier:
The code syntax looks good.

PR #680, opened 2015-04-20
C#7566 - Revisit LP exercise result even if only one attempt is allowed

ilosada:


main/exercice/overview.php
```diff
@@ -155,7 +155,7 @@
     $btn_class = '';
 }
 
-if (!empty($attempts) && $visible_return['value'] == true) {
+if (!empty($attempts)) {
```
ywarnier: I thought this wasn't the solution?

ilosada: This may seem to be a little dramatic change, but all in all the results must be shown depending on the 'results_disabled' test property.

ywarnier: OK, but I expect an explanation about the removal of $visible_return['value'] == true...
If it was removed, then what was it used for before that? (something that avoids the necessity for the reader to review all the code to get to the same conclusion)

ywarnier:
I finally could check this. All good! Thanks.

PR #694, opened 2015-04-30
C#7670 - Fix ticket plugin

ilosada:


plugin/ticket/src/assign_tickets.php
```diff
@@ -13,7 +13,7 @@
 	api_not_allowed();
 }
 $course_info = api_get_course_info();
-$course_code = $course_info['code'];
+$courseId = $course_info['id'];
```
ywarnier: In this case (and the two uses of $courseId below), you should leave it as course_code because the queries on $tblWeeklyReport are queries on a plugin table which maintains the course_code column for now.

PR #699, opened 2015-05-06
B#9413 - Fix advanced subscription modal window

ilosada:


ywarnier:
Angel checked and had other visual issues, so he will be sending another PR with a combined fix.

PR #704, opened 2015-05-12
CT#7535 Bootstrap adjustments in presentation of the css styles exercises tools - Question image

aragonc:


ywarnier:
Missing some images (test-quiz.png in all sizes but 32, and graduation.png in size 16)

PR #736, opened 2015-06-24
Facebook plugin bug correction

baelmyhu:
Do the previous Pull Request then do this one to overwrite files


index.php
```diff
@@ -40,6 +41,11 @@
     });
 </script>';
 
+// Facebook connexion, if activated
+if (api_is_facebook_auth_activated() && !api_get_user_id()) {
```
ywarnier: This should *not* go into index.php, but rather in local.inc.php!

PR #744, opened 2015-06-26
fix $file_name is not defined

mkdgs:


ywarnier:
Hi Michael,
This is great, thanks, but just so you know, we don't develop in master at the moment. I recommend you check branch 1.10.x to contribute to where we are most active right now!

PR #746, opened 2015-06-26
import of multiple answers question with global ponderation are now with negative score by default and adding an extra field in the excel file to configure it - refs BT#8659

NicoDucou:


main/exercice/upload_exercise.php
```diff
@@ -151,6 +151,7 @@ function lp_upload_quiz_action_handling() {
     $feedback_false_index = array();
     $number_questions = 0;
     $question_description_index = array();
+    $noNegativeScore_index = array();
```
ywarnier: Variables MUST be in *all* $camelCaseWithoutDash now :-) (there's another one a bit below)

PR #756, opened 2015-07-03
add a path for custom templates

mkdgs:
add the possibility to override all templates by adding a new one with the same name in /custompages directory
This is useful when you want customize a view and keep all change at the next upgrade


jmontoyaa:
:+1:

PR #771, opened 2015-07-10
Ugly fix, just for the show

mkdgs:
experienced some weird time load issue on shared hosting.
Disabling charset seems to do the trick

### DIFF
--- a/main/inc/lib/document.lib.php
+++ b/main/inc/lib/document.lib.php
@@ -364,16 +364,22 @@ class DocumentManager
             //header('Pragma: no-cache');
             switch ($content_type) {
                 case 'text/html':
+                    $content_type .= '; charset=UTF-8';	
+                    /*
                     $encoding = @api_detect_encoding_html(file_get_contents($full_file_name));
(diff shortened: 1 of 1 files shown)


ywarnier:
I tested the change and it *does* cause issues for me in some cases (with content that is declared as iso-8859-1).
When using the patch: utf8patch1
When not using it: utf8patch2

However, I do recognize that the difference can really be felt.

I suggest we work on that to optimize the api_detect_encoding_html(), which takes in parameter the full HTML document and checks for the existence of <head></head> tags, which I believe can be skipped. We could also only return about 2048 bytes of the HTML file to that function. That should be enough if we don't check for a </head> tag.

ywarnier:
At first glance, reducing to 2048 doesn't seem to have a highly positive impact on performance. I will need to check that in detail with xhprof

ywarnier:
For now, I'm going to add an option in configuration.php to allow for this change to optimize the speed in informed situations.

ywarnier:
Sent in c897c8e180c7bc081ae279a01c07471e066d19d6

PR #817, opened 2015-09-02
Added a CKEditor instead textarea while adding an Event in Agenda - Ref #5208

jloguercio:


jmontoyaa:
Added here:

https://github.com/chamilo/chamilo-lms/commit/509f21560ca9c8a7e11e4846d69044a4f98f707f

PR #822, opened 2015-09-04
Upgrade to WebcamJS from JPEGCam in Document Tools - Refs #7181

jloguercio:


ywarnier:
This PR is OK (reviewed quickly) but we will wait after 1.10.0 to integrate it because it removes the Flash plugin and this might affect IE browsers that do not support WebRTC (or is there a solution for that as well?)

jloguercio:
It doesn't removes Flash plugin, it only uses the the native video api and if it doesn't support that feature it falls to Flash.

Also, in code, you can force to Flash.

ywarnier:
Then it's OK for me.

jloguercio:
image

Tested in IExplorer, works well, it falls to Flash

ywarnier:
Please update your PR to make it against 1.10.x branch. We will add it to 1.10.0 :-)

ywarnier:
Closing as it has already been merged into 1.10.x

PR #838, opened 2015-09-18
Added OpenGraph feature to Social Network and other fixes - Refs #3876

jloguercio:


ywarnier:
Julio, could you confirm that line 1135 of main/inc/lib/social.lib.php is effectively filtered before getting inserted into the database? We believe it is filtered by Database::insert(), but we couldn't find a clear confirmation of that...
If so, could you apply to 1.10.x?

jmontoyaa:
Database::insert cleans already the content.

PR #839, opened 2015-09-21
Remove chat button from elfinder popup file selector - Refs #7841

jloguercio:


main/inc/lib/chat.lib.php
```diff
@@ -244,6 +244,26 @@ public function sanitize($text)
         $text = str_replace("\n", "<br>", $text);
         return $text;
     }
+    
+    public static function  set_disable_chat($status)
```
ywarnier: If this one is static, disable_chat() should be as well, no?
$status could have a default value of "true" considering the function is called "set_disable_chat".
New functions created should be inUpperCaseLetters().
There are *two* spaces between "function" and "set_disable_chat".
There is no documentation block:
/**
 * Disables or enables the chat
 * @param bool $status True to disable, false to enable the chat
 */

main/inc/lib/chat.lib.php
```diff
@@ -244,6 +244,26 @@ public function sanitize($text)
...
+    {
+            $_SESSION['disable_chat'] = $status;
```
ywarnier: Only 4 spaces of indentation.

main/inc/lib/chat.lib.php
```diff
@@ -244,6 +244,26 @@ public function sanitize($text)
...
+    }
+    
+    public function disable_chat()
```
ywarnier: Missing documentation block (and should be static) but then the code using it should change to a static call as well.

main/inc/lib/chat.lib.php
```diff
@@ -244,6 +244,26 @@ public function sanitize($text)
...
+    {
+         if (isset($_SESSION['disable_chat'])) {
+             if (!empty($_SESSION['disable_chat'])){
```
ywarnier: "!empty" already includes "isset", so you don't need to put both. Just if (!empty).
Space missing between )) and {

main/inc/lib/chat.lib.php
```diff
@@ -244,6 +244,26 @@ public function sanitize($text)
...
+                 $status = $_SESSION['disable_chat'];
+                 if ($status == true){
```
ywarnier: Space missing before {

PR #844, opened 2015-09-24
fix a bug when uploading pictures for multi-url

emarguin:
the was a bug there. When a file had already been uploaded, it was impossible to change it (move_uploaded_file does not overwrite old files)


ywarnier:
True... true...

emarguin:
thanks ;) (don't know why netbeans always throughs out my settings...)

PR #856, opened 2015-09-30
Adicional Fix for unsubscribe course (hot courses) - Refs #7850

jloguercio:


main/auth/courses.php
```diff
@@ -156,8 +156,9 @@
 }
 // We are unsubscribing from a course (=Unsubscribe from course).
 if (isset($_GET['unsubscribe'])) {
+    $search_term = isset($_GET['searh_term']) ? intval($_GET['searh_term']) : null;
```
ywarnier: Missing a "c" in "searh_term", no?

jloguercio: Wow, i dont know why that intval() function is there, uhmm there is no "c" variable but this fixes the fact that when you unsubscribe from the main page (index.php - hot courses)  doesn't have a search function there and for that the var search_term doesn't pass for $_GET, only verify if search_term exist

PR #860, opened 2015-10-01
Fix a serious bug that does not allow you to do the exercises

jloguercio:
The function "panelCollapse" its only used in this case (no find another usages)


ywarnier:
I didn't check the feature change but the code looks OK.

jloguercio:
This is the error and doesn't allow you to do the exercise, but with the code change works fine
captura de pantalla de 2015-10-01 11 19 49

PR #875, opened 2015-10-13
add jquery scrollbar and fix document scorm refs - CT#7870

aragonc:


ywarnier:
I sense danger in this one... :jack_o_lantern:

PR #901, opened 2015-10-19
Fix erroneous tag {{course}}

mkdgs:


ywarnier:
These fixes seem to be all covered by the buycourse plugin in version 1.10. However, if you *upgrade*, make sure you check the plugin readme as some database updates have to be applied manually

PR #902, opened 2015-10-21
Fix course image path

mkdgs:


jmontoyaa:
This PR should be also  proposed to 1.10.x

ywarnier:
In 1.10.x, this is treated differently (and already fixed). I just checked.

PR #906, opened 2015-10-26
Fix account creation status, add user extra fields

mkdgs:
Fix account creation status (according to allow_registration setting) 
Add missing user data extra fields


ywarnier:
Could you explain the cases where this is generating a problem (that is fixed by your patch)?

mkdgs:
without this fix:
- When you create an account before in the payment process, this account is not activated even if you have allowed registration without approval. 
- extra user field is not displayed/saved

with this fix:
- the registration works like the regular registration form (the created account need approval only if allow_registration is set to 'approval'.
- extra user field is displayed/saved (it seems to miss extended profile information)

ywarnier:
Merged. Sorry about the delay.

PR #930, opened 2015-11-12
Fixed bug that didn't allow to show the new categories - Refs #7925

jloguercio:


jloguercio:
To see the changes, delete (or ignore) all previous categories created before this fix, this is a insert DB problem, test it adding a new categorie

PR #932, opened 2015-11-13
fix unique_id in event presentation -refs BT#10516

NicoDucou:


main/calendar/agenda.lib.php
```diff
@@ -1428,7 +1428,7 @@ public function get_course_events($start, $end, $courseInfo, $groupId = 0, $sess
                 $user_to_array = $items['users'];
                 $event = array();
                 $event['id'] = 'course_'.$row['id'];
-                $event['unique_id']  = 'course_'.$row['id'].intval($row['session_id']);
+                $event['unique_id']  = 'course_'.$row['id'].intval($row['session_id'].$course_id);
```
ywarnier: Just for the note, it is strange that we have the 'course_'.$row['id'] representing the event ID when the prefix says "course").
In any case, Nicolas, we should NOT join session_id and course_id without a dash sign (_). What happens if you have session 1 and course 22, and then session 12 and course 2? Will that not generate a non-unique ID as well?
Better use 1-22, then 12-2

PR #935, opened 2015-11-13
fix unique_id in event presentation -refs BT#10516

NicoDucou:


ywarnier:
Did you check what would happen if unique_ids were registered with the other method and are now registered with this method? Is there a possibility that this would generate issues for someone with older unique_id ?

NicoDucou:
The unique_id is only used in this function when generating de calendar to avoid double presentaton. It is not saved in the database so there will not be older unique_id to mix with.

PR #939, opened 2015-11-20
not removing '_' and dangerous characters in the database name check …

NicoDucou:
…in step 4 of install form but still checking Mysql special characters required in CT3266 -refs CT7969


ywarnier:
CT#7969 (don't forget the #)

PR #942, opened 2015-11-30
Remove OfficialCode's modification in updateuser since it is not pass…

NicoDucou:
…ed in the parameters - refs BT#10573

### DIFF
--- a/main/webservices/registration.soap.php
+++ b/main/webservices/registration.soap.php
@@ -1916,7 +1916,6 @@ function WSEditUserWithPicture($params)
     $expiration_date = null;
     $expirationDateStatement = '';
     $status = $params['status'];
-    $official_code = '';
     $phone = $params['phone'];
     $picture_url = $params['picture_url'];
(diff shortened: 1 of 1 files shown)


main/webservices/registration.soap.php
```diff
@@ -2014,7 +2013,6 @@ function WSEditUserWithPicture($params)
     $user
         ->setEmail($email)
         ->setStatus($status)
-        ->setOfficialCode($official_code)
```
ywarnier: You should not remove this one. Rather, you should put it in a conditional, like $status, above.

NicoDucou: That's what I did first and then I saw that there is no official_code in the parameters passed to this webservice (tns:editUserWithPicture), so what is the point of adding a check ?

The other option is to add this variable to the parameters, but if I change the declaration of the parameters, it might not work with other system that are using this webservice without this parameter. That's why I decided to delete it.

PR #952, opened 2015-12-15
Add some new info to buy courses payout feature and minor changes - Refs #10614

jloguercio:


plugin/buycourses/lang/spanish.php
```diff
@@ -23,6 +23,9 @@
 $strings['CancelPayout'] = "Cancelar Payout";
 $strings['ContinuePayout'] = "Continuar con el Pago";
 $strings['ProceedPayout'] = "Proceder con el Pago";
+$strings['TheActualPlatformCommissionIs'] = "La actual comisión de la plataforma es de %s, este porcentaje será descontado del total del precio del producto cuya diferencia será el monto base para aplicar estas comisiones.";
```
ywarnier: Should be TheActualPlatformCommissionIsX
The "X" is very important. It is the way we remember (in code) that it goes with a variable

plugin/buycourses/src/payout_report.php
```diff
@@ -70,6 +70,7 @@
         'currency' => $payout['iso_code'],
         'price' => $payout['item_price'],
         'commission' => $payout['commission'],
+        'beneficiary' => $payout['firstname'] . ' ' . $payout['lastname'],
```
ywarnier: Better use api_get_person_name()

PR #972, opened 2016-01-05
Modification of main configuration.dist.php for LDAP

baelmyhu:
Other modification have to be done in translation files whete the process is explained 
because of the new path of files
auth.conf.php => app/config/auth.conf.php
configuration.php => app/config/configuration.php
and the deletion of variable $_configuration\['code_append'\] 
(replaced by "main")


ywarnier:
Changed language variables in French, English and Spanish

PR #973, opened 2016-01-05
Modify app/config/configuration.dist.php for ldap

baelmyhu:
Other changes to do are in translation file because the path of files configuration.php and auth.conf.php has changed.
In configuration.php, path of $extAuthSource has changed too, $_configuration['code_append'] not used anymore.


ywarnier:
It's weird because your PR is considered valid but there is no app/config/configuration.dist.php file in 1.10.x (nor in master). See https://github.com/chamilo/chamilo-lms/tree/1.10.x/app/config
I'm rejecting it because that change is already in main/install/configuration.dist.php, but your change was OK. Apparently I did it manually 5 days after you wrote the commit. Sorry about that.

PR #982, opened 2016-01-30
Variable $ldap_user has not been initialized

AppChecker:
Variable $ldap_user has not been initialized.
Perhaps, varialbe $ldapuser should be used

This possible defect found by AppChecker.


ywarnier:
Thank you for your PR!

PR #990, opened 2016-02-08
Bug fixes on webservices

dnos:
Some bug fixes on webservices:
- Disable user (warning on xml returned when debug mode on)
- Create user (extra fields not created)
- Edit user (extra fields not updated)


ywarnier:
:+1: 
Great! Thank you for your submission.
On a side note, registration.soap.php is much more used and maintained than webservice_user.php, but fair enough.

PR #1009, opened 2016-02-15
Visual Fix when you switch visibility in course home icons

jloguercio:


ywarnier:
Perfect! Please apply to master (if it can be applied)

PR #1044, opened 2016-03-01
Added new style to messaging system in Languages

jloguercio:


ywarnier:
We need to talk about this PR because the "normal-message" class is used by Behat to see if there's an error or not. I need to know if this has changed in other places.

jloguercio:
Nope, this only was changed here, i change it because the normal-message clases has no style (in this case)

PR #1049, opened 2016-03-03
Put message CSS class back into message box to respect Behat testing logic (1.11.x)

jloguercio:


ywarnier:
The commit message is confusing. It should be "Put message CSS class back into message box to respect Behat testing logic"

jloguercio:
Updated !

PR #1054, opened 2016-03-04
Added Feature to customize the logo of each StyleSheet - Refs #8118

jloguercio:


jloguercio:
Found a way to change the logo, without a img cache clean force and db changes

ywarnier:
When I upload an image of the wrong size, I get an error message (that's OK) but the "Current logo" is trying to load http://my.chamilo110.net/web/css/themes/chamilo/images/header-logo-custom.png anyway, so the thumbnail appears broken.

jloguercio:
Updated ! , fix the minor bug for broken thumbnail  !

PR #1075, opened 2016-03-18
Fix FormValidator Css span required visual bug effect - NoRefs

jloguercio:
This fix this visual bug (i can't explain it with words so an image speaks a thousand words)

visualbugformvalidator


jloguercio:
@ywarnier

PR #1082, opened 2016-03-22
Fix UTC problem in session access dates for users - Refs #8143

jloguercio:


ywarnier:
Good for me. Part of the issue might have been to install Chamilo without the date.timezone setting being set, but there was definitely another issue with the UTC time here (fixed now).

PR #1084, opened 2016-03-22
Add config variables for smtp: SMTP_DEBUG and SMTP_SECURE

dnos:
SMTP_SECURE parameter needed if you use Amazon e-mail service, for example. 
Added SMTP_DEBUG in mail config file to enable debug as well.


ywarnier:
Thanks. We're going to release 1.10.4 today, most likely, so after that please send your changes preferably to 1.11.x (you can still send them to 1.10.x, but we will also merge them into 1.11.x and our focus on 1.10.x will be highly diminished).

PR #1092, opened 2016-03-28
Fix CSS Groups View and Message Topics

aragonc:


ywarnier:
This is OK, as far as I can see. I only saw a small missing space...

PR #1105, opened 2016-04-06
Fixed several bugs in 1.11.x

dnos:
Fixed two bugs:
- New group announcement did nothing
- Syntax error in resulttable.class.php, right bracket not closed


dnos:
Added commit to remove link in breadcrumbs to reports main page if user is student boss. The main report page is not showing properly the number of users neither the users when clicking in the links of main/mySpace/index.php

I think this page was meant to be used only for admins, teachers, rrhh resp. but not for student bosses. Although student bosses could access this page by clicking in the breadcrumb.

PR #1180, opened 2016-05-26
apply font-awesome on resource tree (#8256)

mk111:
replaced 0.gif and 1.gif with fa-plus-square-o and fa-minus-square-0, gain enlarging possibility (fa-lg, fa-3x)


mk111:
replaced with #1182, correcting branch

PR #1210, opened 2016-06-02
Add a profile_completed validation to avoid a lot of notices

jloguercio:


jmontoyaa:
This was not needed because you just had to update your DB.

PR #1273, opened 2016-07-06
New feature to let choose which users can see or not which courses in the catalogue

dnos:
This new feature lets set rules in order to define which users can see or not a course in the catalogue.

In order to see a course in the catalogue, the course must comply all other requeriments as before, this has not changed. But additionally it will be possible to define more conditions, the following rules:
- Course A is visible to User X in the catalogue (granted)
- Course B is invisible to User Y in the catalogue (forbidden)

If no rules for a course are defined, then it is shown as before, so it breaks nothing. If some rules are defined then only the users granted can see the course in the catalogue; if defined rules are all to deny access (forbidden) then all users can see the course except those forbidden.

A table has been created to save the rules, and two webservices created to add and remove rules.

TODO: A new screen or tab in course settings to let the administrators choose the users that can see the course in the catalogue, or the users that the course must not be shown. This could be developed in a future release of Chamilo.

The following User Acceptance Tests have been tested:

---- Scenary (Given the following courses and a user X):
 - A: Course without any rule defined
 - B1: Course only with rules of granted permission to see in the catalogue, but not for user X.
 - B2: Course only with rules of granted permission to see in the catalogue, one for the user X. 
 - C1: Course only with rules of forbidden permission to see in the catalogue, but not for user X.
 - C2: Course only with rules of forbidden permission to see in the catalogue, one for the user X.
 - D: Course with rules of granted and forbidden permission to see in the catalogue, but no rule for user X 
 - D1: Course with rules of granted and forbidden permission to see in the catalogue, for the user X just one granting this user to see the course
 - D2: Course with rules of granted and forbidden permission to see in the catalogue, for the user X just one forbidding this user to see the course
 - D3: Course with rules of granted and forbidden permission to see in the catalogue, for the user X one forbidding this user and another granting the same user to see the course

---- UAT (Given that user X could see before all these courses in the catalogue):
- User X CAN see the course A in the catalogue 
(because no rules are defined for the course, everybody can see it)

- User X CANNOT see the course B1 in the catalogue 
(rules are defined for the course, but user is not granted)

- User X CAN see the course B2 in the catalogue
(rules are defined for the course and user is granted)

- User X CAN see the course C1 in the catalogue
(rules are defined for the course but only to forbid, and user is not in the rules)

- User X CANNOT see the course C2 in the catalogue
(rules are defined for the course but only to forbid, and user is in the rules)

- User X CANNOT see the course D in the catalogue
(rules are defined to grant and forbid, but user is not in the rules)

- User X CAN see the course D1 in the catalogue
(rules are defined to grant and forbid, and user is granted in a rule)

- User X CANNOT see the course D2 in the catalogue
(rules are defined to grant and forbid, and user is forbidden in a rule)

- User X CANNOT see the course D3 in the catalogue
(rules are defined to grant and forbid, user is granted but also forbidden in another rule)

### DIFF
--- a/app/Migrations/Schema/V111/Version20160706182000.php
+++ b/app/Migrations/Schema/V111/Version20160706182000.php
@@ -23,7 +23,7 @@ class Version20160706182000 extends AbstractMigrationChamilo
     public function up(Schema $schema)
     {
          $this->addSql(
-             'CREATE TABLE course_rel_user_catalogue (id int(11) NOT NULL AUTO_INCREMENT, user_id int(11) DEFAULT NULL, c_id int(11) DEFAULT NULL, visible int(11) NOT NULL, PRIMARY KEY (id), KEY (user_id), KEY (c_id), CONSTRAINT FOREIGN KEY (c_id) REFERENCES course (id) ON DELETE CASCADE, CONSTRAINT FOREIGN KEY (user_id) REFERENCES user (id) ON DELETE CASCADE) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci'
+             'CREATE TABLE course_rel_user_catalogue (id int NOT NULL AUTO_INCREMENT, user_id int DEFAULT NULL, c_id int DEFAULT NULL, visible int NOT NULL, PRIMARY KEY (id), KEY (user_id), KEY (c_id), CONSTRAINT FOREIGN KEY (c_id) REFERENCES course (id) ON DELETE CASCADE, CONSTRAINT FOREIGN KEY (user_id) REFERENCES user (id) ON DELETE CASCADE) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci'
              );
     }
 
--- a/app/Migrations/Schema/V111/Version20160706182000.php
+++ b/app/Migrations/Schema/V111/Version20160706182000.php
@@ -1,0 +1,42 @@
+<?php
+/* For licensing terms, see /license.txt */
+
+namespace Application\Migrations\Schema\V111;
+
+use Application\Migrations\AbstractMigrationChamilo;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Types\Type;
+
+/**
+ * Class Version20160706182000
+ * Add new table to save user visibility on courses in the catalogue
+ * @package Application\Migrations\Schema\V111
+ */
+class Version20160706182000 extends AbstractMigrationChamilo
+{
+    /**
+     * @param Schema $schema
+     *
+     * @throws \Doctrine\DBAL\DBALException
+     * @throws \Doctrine\DBAL\Schema\SchemaException
+     */
+    public function up(Schema $schema)
+    {
+         $this->addSql(
+             'CREATE TABLE course_rel_user_catalogue (id int(11) NOT NULL AUTO_INCREMENT, user_id int(11) DEFAULT NULL, c_id int(11) DEFAULT NULL, visible int(11) NOT NULL, PRIMARY KEY (id), KEY (user_id), KEY (c_id), CONSTRAINT FOREIGN KEY (c_id) REFERENCES course (id) ON DELETE CASCADE, CONSTRAINT FOREIGN KEY (user_id) REFERENCES user (id) ON DELETE CASCADE) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci'
+             );
+    }
+
+     /**
+      * @param Schema $schema
+      *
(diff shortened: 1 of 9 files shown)


app/Migrations/Schema/V111/Version20160706182000.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * @param Schema $schema
+     *
+     * @throws \Doctrine\DBAL\DBALException
+     * @throws \Doctrine\DBAL\Schema\SchemaException
+     */
+    public function up(Schema $schema)
+    {
+         $this->addSql(
+             'CREATE TABLE course_rel_user_catalogue (id int(11) NOT NULL AUTO_INCREMENT, user_id int(11) DEFAULT NULL, c_id int(11) DEFAULT NULL, visible int(11) NOT NULL, PRIMARY KEY (id), KEY (user_id), KEY (c_id), CONSTRAINT FOREIGN KEY (c_id) REFERENCES course (id) ON DELETE CASCADE, CONSTRAINT FOREIGN KEY (user_id) REFERENCES user (id) ON DELETE CASCADE) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci'
```
ywarnier: Please don't use the size indicator for int types. This is entirely esthetic and doesn't serve any purpose in a PHP context.

main/inc/lib/course_category.lib.php
```diff
@@ -592,13 +592,7 @@ public static function countCoursesInCategory($category_code = '', $searchTerm =
...
         }
 
-        $visibilityCondition = null;
-        $hidePrivate = api_get_setting('course_catalog_hide_private');
-        if ($hidePrivate === 'true') {
-            $courseInfo = api_get_course_info();
-            $courseVisibility = $courseInfo['visibility'];
-            $visibilityCondition = ' AND course.visibility <> 1';
-        }
+    $visibilityCondition = CourseManager::getCourseVisibilitySQLCondition("course");
```
ywarnier: Looks like there's a small identation typo here.

main/inc/lib/course.lib.php
```diff
@@ -4839,14 +5026,8 @@ public static function countAvailableCourses($accessUrlId = null)
             $withoutSpecialCourses = ' AND c.code NOT IN ("' . implode('","', $specialCourseList) . '")';
         }
 
-        $visibilityCondition = null;
+        $visibilityCondition = self::getCourseVisibilitySQLCondition("c");
```
ywarnier: Please use single quotes

main/inc/lib/course.lib.php
```diff
@@ -778,6 +778,108 @@ public static function add_user_to_course($user_id, $courseCode, $status = STUDE
...
+            $cond = array(
+                'user_id = ? AND c_id = ? AND visible = ? ' => array(
+                    $userId,
+                    $courseId,
+                    $visible
+                )
+            );
+            return Database::delete($courseUserTable, $cond);
+        }
+        else {
```
ywarnier: these 2 lines should be on the same line...

main/inc/lib/course.lib.php
```diff
@@ -778,6 +778,108 @@ public static function add_user_to_course($user_id, $courseCode, $status = STUDE
...
+        ];
+        $insertId = Database::insert($courseUserTable, $params);
+
+        return $insertId;
+    }
+
+
+    /**
+     * Remove the user $userId visibility to the course $courseCode in the catalogue.
+     * @author David Nos
```
ywarnier: You should add a reference to contact you here. Either by e-mail or by your github account URL, as the main purpose of specifying the author is to allow other users to contact him in case they have a question about the function ;-)

main/inc/lib/auth.lib.php
```diff
@@ -433,6 +434,8 @@ public function search_courses($search_term, $limit)
             $without_special_courses = ' AND course.code NOT IN (' . implode(',', $special_course_list) . ')';
         }
 
+        $visibilityCondition = ($justVisible?CourseManager::getCourseVisibilitySQLCondition("course"):'');
```
ywarnier: Please use single quotes

ywarnier:
This looks very nice to me (setting apart the few minor observations I sent you). You seem to have caught the way we work and adopted it pretty nicely.

Indeed, it would be good to offer this feature from somewhere else than just a webservice. This should probably remain as an platform-admin-only feature though (I don't think the teacher should decide).

Regarding the user acceptance testing, I think behat is a little broken at the moment, but you should have a look at tests/README.md because behat can help you automate those tests.

@jmontoyaa if you don't have any objection, I'll merge this PR into 1.11.x (just in time before our delayed beta version)

dnos:
Thanks @ywarnier , I have pushed a new commit with the corrections for your observations.

I'll take a look to Behat as well.

PR #1297, opened 2016-07-15
Fixed incorrect method api_get_course_info_by_id

danilobrinu:
Change api_get_course_info_by_id to api_get_course_info -> $courseId is a course code not the index of course

### DIFF
--- a/main/admin/course_user_import.php
+++ b/main/admin/course_user_import.php
@@ -96,7 +96,7 @@ function save_data($users_courses)
 
         if ($_POST['subscribe']) {
             foreach ($to_subscribe as $courseId) {
-                $courseInfo = api_get_course_info_by_id($courseId);
+                $courseInfo = api_get_course_info($courseId);
                 $course_code = $courseInfo['code'];
 
                 if (CourseManager :: course_exists($course_code)) {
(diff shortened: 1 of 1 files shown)


ywarnier:
You spotted an issue, indeed, but your solution isn't complete. The course_rel_user table is now based on the c_id (int) and not the course_code (as it was in 1.9). As such, there is a need for a "translation" of the contents of $csv_subscription to use the c_id instead of the course_code, or to recover the course_code instead of the c_id inside of $db_subscription. At the moment, one array is based on code, the other on ID, so the array_diff() will always return the full array.
So I cannot accept your PR as it is right now, but if you change the course_rel_user query to join on the course table and get the course_code instead of the c_id, I will.

danilobrinu:
Ohh i use 1,10.x and have a issue. I fixed change method api_get_course_info_by_id  to api_get_course_info

jmontoyaa:
I just applied a fix.

ywarnier:
@danilobrinu this should fix the issue you had. If you get a chance to review this and confirm it fixes the problem you had, we would be very glad...

danilobrinu:
Yes @ywarnier. I fixed issue using api_get_course_info insted to api_get_course_info_by_id in my version thx to accepted my PR.

PR #1298, opened 2016-07-15
Fixed incorrect method api_get_course_info_by_id

danilobrinu:
Change api_get_course_info_by_id to api_get_course_info -> $courseId is a course code not the index of course


ywarnier:
See #1297

PR #1390, opened 2016-09-02
Alter inputs dates

csmendonca:
Alter dates inputs for datetime in create new survey, publication and expiration the learning route and in add new datetime in calendar presence


ywarnier:
Assigning to @jmontoyaa for deeper review but seems fine to me.

jmontoyaa:
I think is better to move this to 1.11.x not to master.

PR #1423, opened 2016-09-30
Fix Next or previous button while the item position is first or last - Refs BT#11192

jloguercio:


main/lp/learnpath.class.php
```diff
@@ -11157,9 +11157,9 @@ public function isFirstOrLastItem($currentItemId)
         $lpItemId = [];
 
         $typeListNotToVerify = self::getChapterTypes();
-        foreach ($this->items as $item) {
-            if (!in_array($item->get_type(), $typeListNotToVerify)) {
-                $lpItemId[] = $item->get_id();
+        foreach ($this->get_toc() as $item) {
```
ywarnier: A small comment would be nice (like "//Using get_toc() is equivalent to using $this->items because X and Y")

PR #1481, opened 2016-10-27
Fix duplicate courses in human resources profile

nosolored:
Fixes/Arregla/Corrige #.
- Evita que se muestre en la página user_portal.php cursos duplicados para un usuario con perfil de recursos humanos.


jmontoyaa:
Better to fix the query and add a "distinct" in function SessionManager::get_course_list_by_session_id

nosolored:
That is not resolved with this modification using DISTINCT because there are already previous values in  $personal_course_list [] array

Issue #1532, opened 2016-12-04
Documents: serious inconsistency problems when moving folders


jderriks:
### Expected behavior / Resultado esperado / Résultat attendu
Moving folders and files should check on "move down" or "move to itself"
### Actual behavior / Resultado real / Résultat réel
Folders are remover=d from filesystem but are still in database. Errors when trying to open them and errors like "inconsistency detected" when trying to delete.
### Steps to reproduce / Pasos para reproducir / Étapes pour reproduire
Many cases that lead to disaster (for teachers that don't keep local backups)
Testcase 1: Moving a folder in documents to the same document folder will DELETE the folder from disk but not from the database. This is the simplest test.
Testcase 2: Create a folder structure documents/a/b/c and try moving documents/a to documents/a/b This will crash the website and recursively create a deep nested file structure on disk (filesystem limit). 

There were more bugs with the document MOVE. I think I fixed them in the pull request.

If you want to know if your system is OK, check error-log for errors like:
_"/main/inc/lib/document.lib.php 1322: System inconsistency detected. The file or directory 
/chamilo-lms/app/courses/TEST9/document/f2 seems to have been removed from the filesystem independently from the web platform. To restore consistency, the elements using the same pa
th will be removed from the database_"

Users will mostly see errors like this:
<img width="954" alt="screen shot 2016-12-04 at 17 39 32" src="https://cloud.githubusercontent.com/assets/1659604/20867593/c04d9f26-ba48-11e6-9c37-feed08c1cba0.png">

jmontoyaa:
Indeed that correction fix the problem. Can you provide a PR?

jderriks:
https://github.com/chamilo/chamilo-lms/pulls

jmontoyaa:
Merged.

ywarnier:
I suppose we all agree to close this one, then?

jderriks:
If https://github.com/chamilo/chamilo-lms/pull/1540 is pulled it will work on Windows as well.

ywarnier:
Oh, yes, sorry, I mistakenly thought it had already been pulled. Reviewing it, I realize something might be missing. Could you review my comment?

jderriks:
It seems to work now. A fsck-like tool to check for consistency filesystem vs database would be a nice tool though.

PR #1533, opened 2016-12-04
Github #1532 Filesystem inconsistency when moving folders

jderriks:
Needs testing on student work (assignment) features that also uses the move function.

I changed default setting to use the OS exec("mv") function because the symfony filesystem functions are buggy and slow.

Needs testing on WINDOWS too.


jmontoyaa:
Remember to update the variables here: 
http://translate.chamilo.org/ 
Otherwise your variables will be overwritten after a global update in the next release.

PR #1540, opened 2016-12-08
 Github #1532 followup fix for WindowsOS case

jderriks:
Extra fixes:
-check if file already exists on targetdir to prevent duplicate names in database and GUI (but not in filesystem)
-if isWindowsOS is true, copyToDir is used but then a directory must be created first.

### DIFF
--- a/main/inc/lib/fileManage.lib.php
+++ b/main/inc/lib/fileManage.lib.php
@@ -192,6 +192,9 @@ function move($source, $target, $forceMove = true, $moveContent = false)
 
         /* File case */
         if (is_file($source)) {
+            if (is_file($target . '/' . $file_name)) {
+                return false; // target file already exists
+            }
             if ($forceMove && !$isWindowsOS && $canExec) {
                 exec('mv ' . $source . ' ' . $target . '/' . $file_name);
             } else {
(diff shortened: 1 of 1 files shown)


main/inc/lib/fileManage.lib.php
```diff
@@ -192,6 +192,9 @@ function move($source, $target, $forceMove = true, $moveContent = false)
 
         /* File case */
         if (is_file($source)) {
+            if (is_file($target . '/' . $file_name)) {
```
ywarnier: I think we should also take the $forceMove param into account: $forceMove==true should imply we move the file even if the destination already exists. What do you think?

jderriks: It is the !$isWindowsOS condition that causes Windows system not to use the exec method because rm and mv are linux commands. 
ForceMove=true for Linux is fine and it now checks if a file or folder allready exists.

jmontoyaa:
I added a fix for that.

ywarnier:
@jmontoyaa not sure what must be tested here and if we still need to include this PR or if we can close it. Please advise.

jmontoyaa:
I already added a fix for that. To test this case Chamilo should be running in a windows server and try to move folders in the Document tool.

ywarnier:
@jderriks just a last check: this works for you now on Windows (with the changes already included by @jmontoyaa), right?

jderriks:
works.

PR #1560, opened 2016-12-19
Error with query construction, missing "( )"

Starbugstone:
error when the user search string in auth.conf.php for the LDAP search to import users

adding the extra "( )" englobes the full query and renders it correct as a LDAP filter.


ywarnier:
Looks good to me. I can't see in which context this would make it worst either.
Thanks!

PR #1596, opened 2017-01-19
Update trad4all.inc.php

simbulu:
These are the currently translated strings we use in the Faroe Islands.
The focus has been  on translating whatever the students see doing exercises / quizzes. 
Note that we removed the trailing '?>' on purpose, as anything, e.g. newline, trailing such a marker gives serious problems, like invalid image files.


jmontoyaa:
Thanks, I just updated the lang variables

PR #1619, opened 2017-02-02
Change session name from 50 to 150 chars in session edit - BT#12232

jloguercio:


jmontoyaa:
This should be a constant and added in api.lib.php

PR #1623, opened 2017-02-03
fix Button subscribe to session

aragonc:


main/auth/courses_controller.php
```diff
@@ -518,8 +518,14 @@ public function getRegisteredInSessionButton(
         $sessionId,
         $sessionName,
         $checkRequirements = false,
-        $includeText = false
+        $includeText = false,
+        $btnBing = false
     ) {
+        if($btnBing){
```
AngelFQC: PSR http://www.php-fig.org/psr/psr-2/#if-elseif-else :smile:

PR #1634, opened 2017-02-08
Added session course users and session users export button to resume session - Refs BT#12221

jloguercio:


jmontoyaa:
Both code is almost the same. 
main/admin/user_export.php and 
main/user/user_export.php 

There should be only one file that does the job.

PR #1689, opened 2017-02-17
Fix undefined index

hi-rafa:
**While I was testing the edition of a User profile in developer mode I found this small error:**

screen shot 2017-02-16 at 4 42 11 pm

**Before I fixed it I checked where the variable was coming from and I believe it was supposed to be a variable from and input. I did check the view but I couldn't find any input with the name address.**

screen shot 2017-02-17 at 8 45 23 am

# **Could it be that the variable $user['address'] is obsolete?**


jmontoyaa:
:+1:

PR #1701, opened 2017-02-20
Fix DateRangePicker that doesn't allow to pass the Required Rule validation - Refs BT#12317

jloguercio:


main/inc/lib/formvalidator/Element/DateRangePicker.php
```diff
@@ -194,5 +195,7 @@ public function getSubmitValue($value, &$submitValues, &$errors)
         }
         $submitValues[$elementName.'_start'] = $parsedDates['start'];
         $submitValues[$elementName.'_end'] = $parsedDates['end'];
+
+        return $value;
```
jloguercio: Yep ... this is it ...

Issue #1725, opened 2017-03-06
pChart breaks with PHP 7.1


ywarnier:
### Current behavior / Resultado actual / Résultat actuel
If using PHP 7.1, pChart breaks on a series of fatal errors. No chart is getting generated.
Reason? A particular change in PHP 7.1 that prevents the use of the empty index on a string (which was previously silently converted to array):
http://php.net/manual/en/migration71.incompatible.php#migration71.incompatible.empty-string-index-operator
The main issue here is that pChart is not being developed since January 2014. c-pchart (https://github.com/szymach/c-pchart) is not being maintained since January this year. It is time to either contribute PRs to this one or to fork it.
I have created an issue there to see what the author thinks: https://github.com/szymach/c-pchart/issues/25

### Expected behavior / Resultado esperado / Résultat attendu
No error

### Steps to reproduce / Pasos para reproducir / Étapes pour reproduire
Use PHP 7.1
Load the first page of a session, as an admin (or anyone).
You get something like this:
Fatal error: Uncaught Error: Cannot use string offset as an array in /var/www/chamilo111x/vendor/szymach/c-pchart/src/Classes/pData.php on line 44

### Chamilo Version / Versión de Chamilo / Version de Chamilo
1.11.x

jmontoyaa:
"the first page of a session" ? Can you provide an URL?
I have php 7.1.1 installed so I can test this one.

ywarnier:
/main/session/index.php?session_id=1

jmontoyaa:
Works for me ...

![selection_006](https://cloud.githubusercontent.com/assets/158935/23658909/91d1015c-0344-11e7-9aa2-ef540f4d5756.png)

I have this PHP version and extensions:

```
php -v
PHP 7.1.1-1 (cli) (built: Jan 25 2017 14:31:59) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.1.1-1, Copyright (c) 1999-2017, by Zend Technologies

```

```
php -a
Interactive mode enabled

php > print_r(get_loaded_extensions());
Array
(
    [0] => Core
    [1] => date
    [2] => libxml
    [3] => openssl
    [4] => pcre
    [5] => zlib
    [6] => filter
    [7] => hash
    [8] => pcntl
    [9] => Reflection
    [10] => SPL
    [11] => session
    [12] => standard
[...]

PR #1727, opened 2017-03-06
Use pChart 2.0.3 to support PHP 7.1 - refs #1725

ywarnier:
This PR fixes the support for PHP 7.1. However, it also removes support for PHP 5.4, of which the development has stopped in 2014 and support has stopped in 2015. So not sure we could not boost the requirement a bit here and move it to PHP 5.5...
@jmontoyaa any opinion on the matter?

In a sense, with this change we have to decide: either we support 7.1 or we support 5.4, but not both at the same time.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,7 +75,7 @@
         "zendframework/zend-http": "2.5.1",
 
         "ezyang/htmlpurifier": "4.6.0",
-        "szymach/c-pchart": "2.0.3",
+        "szymach/c-pchart": "2.0.4",
         "aferrandini/phpqrcode": "1.0.1",
         "mpdf/mpdf": "6.1.*",
         "studio-42/elfinder": "2.1.*",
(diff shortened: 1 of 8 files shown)


jmontoyaa:
Looks legit, but 1.11.x requires php 5.4 and that package requires php 5.5.

jmontoyaa:
But what about chamilo installations with php 5.4?

ywarnier:
Yes, that's the question. Or better said, the question is "Do we still want to support 5.4 for showing the charts?". I guess we'll have to leave this PR open until we move to 2.0, and know that it's there if we have users using PHP 7.1 and their charts don't work.

ywarnier:
I discussed this with c-pchart's author here https://github.com/szymach/c-pchart/issues/27 and he's downscaled the PHP requirement to PHP 5.4 in version 2.0.4 (https://github.com/szymach/c-pchart/pull/28), so there's no need to delay this PR anymore.

jmontoyaa:
:open_mouth: Great!!

PR #1730, opened 2017-03-08
Fix lost password require file path

hi-rafa:
## Error
error1
error2
error3
error4
error5

## How to reproduce error
1 - Change column "selected_value" to 'true' table settings_current where 'variable' = use_custom_pages.
2 - Copy and rename custompages/index-unlogged-dist.php to custompages/index-unlogged.php

### Current behavior / Resultado actual / Résultat actuel
After you send the password recovery form, the program includes the file (index-unlogged.php) rather than redirecting to home page., so the url stays like this (http://chamilo.loc/main/auth/lostPassword.php).

### Expected behavior / Resultado esperado / Résultat attendu
After the form is sent the program should redirect you to the home page and display a success message

screen shot 2017-03-08 at 11 11 31 am

### Why chamilo don't set the full file path in the includes?

http://stackoverflow.com/questions/8354362/what-is-the-best-practice-for-including-php-files


jmontoyaa:
Nice!!   :smile: :sunglasses:

hi-rafa:
Hi @jmontoyaa 
What should with the other issues I mentioned above?

jmontoyaa:
> Why in the file lostPassword.php it includes files after the form is sent?

In lostPassword.php there's only one require:
https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/auth/lostPassword.php#L18

If you mean the custom pages code, this was developed (by cblue) to change the page layout after you submit the form  using the custompages/lostpassword.php file. You have to develop your own pages with the structure you want using the custompages/*.dist.php files. Those files are just an example.

> Why chamilo don't set the full file path in the includes?

if you're talking about "require_once('language.php');" yes it should be fixed.

hi-rafa:
### Why in the file lostPassword.php it includes files after the form is sent?

screen shot 2017-03-10 at 9 18 42 am



### Why chamilo don't set the full file path in the includes?

screen shot 2017-03-10 at 9 11 41 am

jmontoyaa:
> **This includes creates the bug I mentioned above

This is because after the form is validated it will be redirected to the custom page.

> Why chamilo don't set the full file path in the includes?

You're right the files should be called using api_get_path(LIBRARY_PATH) or other constant depending the location of the file, except the global.inc.php that uses the __DIR__ constant.

hi-rafa:
Thanks for getting back to me so quickly. 
I know you guys have more important sections to develop I just wanted to let you know that I think there is still some work to be done in the display custom pages system.

74e2b320-03f7-11e7-9f6e-630889f7a43b

PR #1739, opened 2017-03-11
Fixes a visual bug with radio buttons when selecting prerequisites.

SatyanJ:
The label was misplaced. Radiobutton was on top of the icone and not
aligned with the 'none' option.

### Before:
before
### After:
after

(@moy)


moy:
Sounds good to me.

PR #1766, opened 2017-03-20
Feature optional rating

hi-rafa:
screen shot 2017-03-20 at 3 08 03 pm

screen shot 2017-03-20 at 3 08 24 pm

screen shot 2017-03-20 at 3 07 45 pm

screen shot 2017-03-20 at 3 08 31 pm

screen shot 2017-03-20 at 3 13 40 pm

screen shot 2017-03-20 at 3 17 18 pm


hi-rafa:
Hi @ywarnier,
I need to know if chamilo wants this feature, because if chamilo doesn't want it I will change the implementation and hide the votes with css.

Thanks

ywarnier:
Hi @hi-rafa,
Yes, I like the option. However, the min-height should remain (even if reduced a bit), otherwise we will get issue when titles or lists of teachers are a bit larger. So I would say: just reduce the min-height from the amount of px you save by removing the votes, but do not delete it.

jmontoyaa:
I added the changes without the css. You can add that css change in your custom theme.

ywarnier:
@hi-rafa if you're OK with the partial integration of @jmontoyaa, let us know (if you're not, let us know as well) so we can close this PR.

hi-rafa:
Hi @ywarnier,
For me its fine to not commit the css changes, but how to notify or warn the developers that if they activate this option they should also fix the interface?

ywarnier:
@jmontoyaa  please don't mix code styling and code changes in a single commit... :-/ (impossible to review quickly)

jmontoyaa:
Oops my fault. I think this is not the right commit.

jmontoyaa:
Somehow this is a merge. In my commit I have only 2 files:

courses_categories.php and
configuration.dist.php

I think the "amend commit" option in phpstorm does this "magic stuff".

jmontoyaa:
selection_011

ywarnier:
@hi-rafa I think the reason why @jmontoyaa didn't include your CSS changes is because you sent them to web/css/ while you should send them to app/Resources/public/css/
web/css is dynamically generated by "composer install", while app/Resources/public/css/ is the permanent code of Chamilo CSS.
If you want to send your change again, to the right place, we can have a look, although as far as I can see on your screenshot it's mostly fixing a min-height, which I'm not sure will fix the current style issue (too much white space).

ywarnier:
@jmontoyaa ok, thanks for the analysis. I was beginning to think you were doing it on purpose :-p

jmontoyaa:
Updated

hi-rafa:
Hi @ywarnier  & @jmontoyaa 
What happened to this pull request? Why isn't merged?

jmontoyaa:
The code was added:
It doesn't work when adding this?

`$_configuration['hide_course_rating'] = true;`

hi-rafa:
Hi @jmontoyaa 
Forgive I just check the chamilo 1.11.x and the code its been merged. 
My bad, I though the pull request wasn't merged.

Issue #1754, opened 2017-03-15
Bug when restoring only quiz with questions categories and images from a backup


baelmyhu:
Questions with categories not restored.
Image file used in questions not restored.

The correction solve the question with category problem.
It restores images needed in questions.
It restores questions categories.

---
Ivan has developped the correction he's going to submit.

PR #1771, opened 2017-03-21
Bug when restoring only quiz with questions categories and images from a backup Issue #1754

pielRouge:
Solution for issue #1754


jmontoyaa:
I added code based in this PR
Use of hotspot with quiz-"time".jpg and quiz hotspot copy is fixed.
I also added the course id in the test category.

But I didn't add the part of CourseSelectForm.php not sure what's the idea of that.

ywarnier:
@pielRouge could you attend @jmontoyaa 's question about CourseSelectForm.php? We need more info. If the current partial merges by @jmontoyaa are OK for you, then we could close this PR...
I will need an answer within 7 days (otherwise I will close it).

pielRouge:
The idea of CourseSelectForm.php is about recover and restore the images of the quiz. Currently if an user selects one quiz to be restore, he have to select the images and the test category too if he wants that his quiz works in another course. We think that is more easy for the user if he selects the quiz and the selection of his image and his category is made automatically. 

The user have to select just his quiz and we recover  and restore the images and the category at the same time. 

Otherwise I'm OK with the current partial merges by @jmontoyaa

ywarnier:
@pielRouge to avoid creating a mess or leaving this PR open forever, would you be so kind as to send the CourseSelectForm part as another pull request?

pielRouge:
@ywarnier Of course. I will do it.

PR #1773, opened 2017-03-21
1.11.x

aragonc:


app/Resources/public/css/base.css
```diff
@@ -6175,6 +6182,15 @@ ul#toolnavbox-two li a.btn{
     .installer{
         display: inline-table;
     }
+    .header-movil{
```
AngelFQC: http://getbootstrap.com/css/#responsive-utilities-classes :stuck_out_tongue_winking_eye:

PR #1809, opened 2017-04-05
fix message alert

aragonc:


jmontoyaa:
Maybe is better to just use Display::addFlash() to prompt a message ?

PR #1885, opened 2017-04-27
Implementation of api_get_language_isocode (2.0)

SatyanJ:
The function wasn't implemented and returned the locale language instead of the iso code. 

PS: I failed and base the correction on #1882 branch instead of master.

(cc @moy)


SatyanJ:
I'll cherry-pick this fix on the clean master. And do a PR.

Issue #1891, opened 2017-04-28
Undefined index when clicking on a course (2.0)


sjcqs:
### Current behavior / Resultado actual / Résultat actuel
You cannot see a course. An error is thrown `Undefined index 'english'`. This error is due to the fact that api_get_language_isocode is not implemented and returns 'english'
### Expected behavior / Resultado esperado / Résultat attendu
Course page should appear.
### Steps to reproduce / Pasos para reproducir / Étapes pour reproduire
Click on a course with a teacher.
### Chamilo Version / Versión de Chamilo / Version de Chamilo
2.0

(A fix is incoming ) 

(cc @moy)

jmontoyaa:
I added different fixes for this ... To check

ywarnier:
No feedback provided. Closing.

PR #1892, opened 2017-04-28
Implementation of api_get_language_isocode

SatyanJ:
`api_get_interface_language` is deprecated and replace with `api_get_language_isocode`.
This function shall return the isocode of the platform language.
(fix for #1891)

(cc @moy)


jmontoyaa:
There's no need to do that change.

`Container::getTranslator()->getLocale();`

That code is in charge to return the locale of the platform.
The locale value is changed depending on on the listener here:
`
src/Chamilo/CoreBundle/EventListener/LocaleListener.php`

This is the place where the locale changes depending some parameters (user/platform/course/etc).

SatyanJ:
I got an error when trying to display a course, `Container::getTranslator()->getLocale();` returns 'english' instead of 'en'.

jmontoyaa:
That's then an error.

In which page, URL? 

I have currently "en" when entering to:

"web/app_dev.php/main/admin/course_list.php" or
"web/app_dev.php/userportal"

Did you check that you have correctly the locale in:

`app/config/parameters.yml`

SatyanJ:
It might be link to the use of the course filler. When I use it and click on a generated course there an error but with a manually created course it's ok.

jmontoyaa:
I just added a fix for the course filler.

Issue #1694, opened 2017-02-20
Error in the creation of a new Learning path


hi-rafa:
### Current behavior / Resultado actual / Résultat actuel
Php error in the creation of a new Learning path

<img width="1160" alt="1" src="https://cloud.githubusercontent.com/assets/3813277/23118621/1d1d3b78-f755-11e6-9d5f-2b80da8f0f70.png">
<img width="1223" alt="2" src="https://cloud.githubusercontent.com/assets/3813277/23118625/1f2dae34-f755-11e6-8e95-a2cf44ff515f.png">
<img width="1186" alt="3" src="https://cloud.githubusercontent.com/assets/3813277/23118633/20d00c64-f755-11e6-8691-4e3ee966c33a.png">

### Expected behavior / Resultado esperado / Résultat attendu
Step 1 Learning path

### Steps to reproduce / Pasos para reproducir / Étapes pour reproduire
admin > courses > selected course > Learning path > new Learning path

### Chamilo Version / Versión de Chamilo / Version de Chamilo
1.11.x

jmontoyaa:
You need to update chamilo with the latest code in 1.11.x 
Also execute a composer update

hi-rafa:
<img width="1251" alt="screen shot 2017-02-22 at 9 31 41 am" src="https://cloud.githubusercontent.com/assets/3813277/23203232/f99c321a-f8e1-11e6-94f1-3b5cf434c174.png">
![screen shot 2017-02-22 at 9 29 05 am 2](https://cloud.githubusercontent.com/assets/3813277/23203233/fb11d000-f8e1-11e6-938f-30b5c01fef5c.png)
<img width="1275" alt="screen shot 2017-02-22 at 9 29 15 am" src="https://cloud.githubusercontent.com/assets/3813277/23203237/fc7c27e2-f8e1-11e6-8fab-946315aea48b.png">


Error still occurs.

jmontoyaa:
Check if you have this file:
`
main/inc/lib/formvalidator/Element/DatePicker.php`

Check if the file is correctly added in the autoloader here:

vendor/composer/autoload_classmap.php

`'DatePicker' => $baseDir . '/main/inc/lib/formvalidator/Element/DatePicker.php',`

jmontoyaa:
try also doing a git checkout of 1.11.x

ywarnier:
[...]

PR #1897, opened 2017-04-29
DatePicker.php was deleted by composer update.

jderriks:
https://github.com/chamilo/chamilo-lms/issues/1694
Every time you do a composer update, the DatePicker.php was removed and not added to the autoloader files generated by 
` composer dump-autoload -o`
(on Mac and windows php unlink will remove DatePicker.php even when datepicker.php is written lowercase)

### DIFF
--- a/main/announcements/announcements.php
+++ b/main/announcements/announcements.php
@@ -485,7 +485,7 @@ switch ($action) {
 
         $defaults['email_ann'] = true;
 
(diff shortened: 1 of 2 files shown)


jmontoyaa:
This happens only windows systems. I added the fix

jderriks:
Not only windows. On Mac's the default filesystem is case-insensitive too.
Thanks for the fix!

PR #1906, opened 2017-05-02
Accessibility improvement

Polycapa:
Resume

- aria-label were added to inputs without label
- form were correctly linked to their corresponding labels

(cc @moy)


main/admin/course_add.php
```diff
@@ -42,7 +42,14 @@
 $form->addElement('header', $tool_name);
 
 // Title
-$form->addText('title', get_lang('Title'), true);
+$form->addText(
+    'title',
+    get_lang('Title'),
+    true,
+    array(
+        'id' => 'title'
```
ywarnier: It seems dangerous to me to add an HTML ID on a page (any page) as "title". This is just a very common word and just a small lack of awareness could make you create another element with the same ID. Please change to "course-title" or something like that.

main/admin/settings.lib.php
```diff
@@ -1327,7 +1327,7 @@ function generateSettingsForm($settings, $settings_by_access_list)
                             get_lang($row['comment']),
                             get_lang('MB')
                         ),
-                        array('maxlength' => '8')
+                        array('maxlength' => '8', 'id' => $row['variable'])
```
ywarnier: This is slightly dangerous too. I would rather build a string like 'id' => 'setting-'.$row['variable']
There are a few settings that can be added through plugins and it would be better to avoid any kind of weird circumstance where the ID conflicts with another basic element of the page.

main/admin/settings.lib.php
```diff
@@ -1349,6 +1349,7 @@ function generateSettingsForm($settings, $settings_by_access_list)
                     continue;
                 } else {
                     $hideme['class'] = 'col-md-4';
+                    $hideme['id'] = $row['variable'];
```
ywarnier: Same here, I guess (and a few other cases below, in this same file)

main/admin/user_list.php
```diff
@@ -816,7 +816,14 @@ function status_filter($status) {
 
 // Create a search-box
 $form = new FormValidator('search_simple', 'get', null, null, null, 'inline');
-$form->addText('keyword', get_lang('Search'), false);
+$form->addText(
+    'keyword',
+    get_lang('Search'),
+    false,
+    array(
+        'aria-label' => get_lang("SearchUsers")
```
ywarnier: And here, for some reason, you don't add an ID :-)

main/inc/lib/sessionmanager.lib.php
```diff
@@ -7131,7 +7131,7 @@ public static function setForm(FormValidator $form, array $sessionInfo = [])
             'name',
             get_lang('SessionName'),
             true,
-            ['maxlength' => 150]
+            ['maxlength' => 150, 'id' => 'name']
```
ywarnier: Same as "title". Please avoid IDs that are very generic English words.

main/admin/settings.php
```diff
@@ -80,7 +80,9 @@
     array(),
     FormValidator::LAYOUT_INLINE
 );
-$form_search->addElement('text', 'search_field');
+$form_search->addElement('text', 'search_field', null, array(
+    'id' => 'search_field'
+));
```
AngelFQC: A `aria-label` is needed here 😄

main/admin/settings.lib.php
```diff
@@ -1369,18 +1370,37 @@ function generateSettingsForm($settings, $settings_by_access_list)
                     if (file_exists($file)) {
                         $value = file_get_contents($file);
                     }
-                    $form->addElement('textarea', $row['variable'], array(get_lang($row['title']), get_lang($row['comment'])) , array('rows'=>'10'), $hideme);
+                    $form->addElement(
+                        'textarea',
+                        $row['variable'],
+                        array(get_lang($row['title']), get_lang($row['comment'])),
+                        array('rows'=>'10', 'id' => $row['variable']),
```
AngelFQC: Spaces before and after of `=>`

main/admin/settings.lib.php
```diff
@@ -1369,18 +1370,37 @@ function generateSettingsForm($settings, $settings_by_access_list)
...
                     $value = '';
                     if (file_exists($file)) {
                         $value = file_get_contents($file);
                     }
-                    $form->addElement('textarea', $row['variable'], array(get_lang($row['title']), get_lang($row['comment'])) , array('rows'=>'10'), $hideme);
+                    $form->addElement(
+                        'textarea',
+                        $row['variable'],
+                        array(get_lang($row['title']), get_lang($row['comment'])),
+                        array('rows'=>'10', 'id' => $row['variable']),
```
AngelFQC: ` => `

main/admin/settings.lib.php
```diff
@@ -1369,18 +1370,37 @@ function generateSettingsForm($settings, $settings_by_access_list)
...
+                    );
                     $default_values[$row['variable']] = $value;
                 } else {
-                    $form->addElement('textarea', $row['variable'], array(get_lang($row['title']), get_lang($row['comment'])) , array('rows'=>'10'), $hideme);
+                    $form->addElement(
+                        'textarea',
+                        $row['variable'],
+                        array(get_lang($row['title']),
+                        get_lang($row['comment'])),
+                        array('rows'=>'10', 'id' => $row['variable']),
```
AngelFQC: And here

main/admin/course_add.php
AngelFQC: I'm adding an `id` by default (if not is set) when a element is added with `FormValidator`. The `id` for elements will be `the name of form + underscore + the element name`. https://github.com/chamilo/chamilo-lms/commit/013347ff73167cca86fe9ffe3dd11a972bbfd8da

This should improve the use of the `for` attribute in the `label` tags (At least with the screen readers)

AngelFQC:
@Polycapa please, pull the last changes from chamilo 1.11.x branch and resolve the conflicts with your code. Keep in mind this commit https://github.com/chamilo/chamilo-lms/commit/013347ff73167cca86fe9ffe3dd11a972bbfd8da

main/admin/course_add.php
```diff
@@ -47,7 +47,7 @@
     get_lang('Title'),
     true,
     array(
-        'id' => 'title'
+        'aria-label' => get_lang('Title')
```
AngelFQC: Now, in some cases, is not necessary add an id or aria-label (please update your repo)

jmontoyaa:
Instead of adding "aria-label" in all the places I think it should be better to add one time in FormValidator.
The same for others attributes, title, alt, etc...

For **example** in the case of text input:

```
 public function addText($name, $label, $required = true, $attributes = array())
 {
     $attributes['aria-label']  => $label];
     $element = $this->addElement('text', $name, $label, $attributes);
```

AngelFQC:
Maybe, but I understand than `aria-label` is needed when the input field has not label.
And the way to assign a label to a input field is by `<label for="input_id">`

See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

jmontoyaa:
Yes, the idea will be that FormValidator handles aria-label.

PR #1907, opened 2017-05-02
Feature add field viewed_at

miky96es:
Add field in gradebook certificate table, which saves datetime about when the user see for first time the gradebook certificate.

TODO: 
Implement code in /certificates/index.php that saves this mentioned info.


jmontoyaa:
You need also to change the entity, and add a variable setting the name and type and if nullable or not, and also adding the "set" and "get".

`src/Chamilo/CoreBundle/Entity/GradebookCategory.php`

Issue #2146, opened 2017-09-16
1.11.4 Bug on group creation when group categories are enabled 


mk111:
applies to  1.11.4 /1.11.6

situation: 
categories for groups are enabled
table c_group_category contains single record : for Default group with id=2

after creating more than one group (for eq.: 3)  based on Default **with same properties** (same cat, same seats), groups are created without problem. Visualization of the new groups shows there are TWO categories, both named DEFAULT, first containing single group, second other two groups.

So virtually there are 2 groups with the same name.

I'm attaching the content of c_group_info table: 
category id=2 is set only for the first group, other got id=0,and there is a difference in other attributes, so i assume there is a problem with checkbox-es (same for all groups)
![group](https://user-images.githubusercontent.com/4478116/30516415-f59515ce-9b3e-11e7-99c4-73d951ef5d98.png)

user workaround : edit misplaced group name (add space at the end and save). Group will be placed to the right place/category. If you repeat the procedure for every misplaced group, now empty category will disappear when 1.11.6, but will be shown empty when using 1.11.4

change proposal:
https://github.com/chamilo/chamilo-lms/pull/2147

ywarnier:
Not sure we're going to be able to review this properly before publishing 1.11.6, but we'll try.

ywarnier:
@mk111 could you check whether this seems fixed to you? If not, then we'll consider it fixed next Friday afternoon. Thanks for reporting!

mk111:
tested,; works fine.

PR #2147, opened 2017-09-17
fix group id assigning

mk111:
when creating groups within category with same category option selected


jmontoyaa:
Closing should be fixed in #2146

Issue #2222, opened 2017-11-21
Announcements: required fields cannot be resubmitted 


jderriks:
### Current behavior / Resultado actual / Résultat actuel
You cannot send an announcement after you tried to sending one with a missing required element.
### Expected behavior / Resultado esperado / Résultat attendu
Just send it if fields are ok.
### Steps to reproduce / Pasos para reproducir / Étapes pour reproduire

- Create an announcement and leave only the subject empty (required field)
- click send, Form will be returned with notice: required field.
- fill in the subject and click send
- Form will re-appear without error and NOT sending announcement
- and now you are in a loop, form cannot be send.

This is caused by the "token security" checks. Quick fix in announcements.php:
+++ b/main/announcements/announcements.php
@@ -23,7 +23,10 @@ api_protect_course_script(true);
 api_protect_course_group(GroupManager::GROUP_TOOL_ANNOUNCEMENT);

 $ctok = Security::get_existing_token();
-$stok = Security::get_token();
+//if form not valid because of missing required fields, dont change token
+if ($ctok != $_POST['sec_token']){
+    $stok = Security::get_token();
+}

 
### Chamilo Version / Versión de Chamilo / Version de Chamilo
Git branch 1.11.x nov 20 2017
but you can also test it on https://11.chamilo.org/index.php

jderriks:
The TO: and FROM: select box should keep the names sorted. Now they go all over the place if you move them from left to right and vv.
I created a PR:
https://github.com/chamilo/chamilo-lms/pull/2224

jmontoyaa:
Should be fixed with:

https://github.com/chamilo/chamilo-lms/commit/e41c5f824e97579096f25ace1deceef9eac76eee

ywarnier:
@jderriks could you let us know by next Monday if that works for you?

jderriks:
I don't know how to test the original purpose of the security token (some XSS block?).
The resubmission of the announcement now works fine.
The send-to-all issue still remains: too many times a teacher sends a message to all users in a course because "send to all"  is the default and that is an UX issue. In older versions of Dokeos there used to be a popup "are you sure" but I think my ":ALL" group fix makes the UI more consistent: allways select a group or recipient and then you can remove the instruction "if you do not select a user you will sent spam to all users in the course" which irritated a lot of students and teachers at our university.
Default "doing nothing" behaviour should be as harmless as possible.

jmontoyaa:
[...]

PR #2224, opened 2017-11-21
Fix several announcement issues like #2222 and more

jderriks:
Prevent Teachers Spam. Select ALL is you want to mail everyone.


ywarnier:
@jmontoyaa I don't see that your commit solves the issue mentioned in #2226 about the order of users in the left list.

ywarnier:
This PR, however, does fix the issue. Please review and merge.

jmontoyaa:
@ywarnier The problem with this PR is that it was added for 2 different issues.
The token issue was fixed. I cannot reproduce the error.

jmontoyaa:
I just added a fix for that select ordering.

ywarnier:
Now I'm good. Thanks.

PR #2228, opened 2017-11-27
fix #2125 forum thread order 0 default sticky value

### LINKED ISSUE #2125: Forum: Thread ordering changes after edit
see PR the cause was that DB 'sticky' field is default 0 but an edit changes it to NULL. That's a real bug.

## Current behavior / Resultado actual / Résultat actuel
If you have a number of threads and you edit the *body* of the first post it "drops down" in the list. Why?
### Expected behavior / Resultado esperado / Résultat attendu
The Thread ordering should be based on name or time of creation but not random after edit.
### Steps to reproduce / Pasos para reproducir / Étapes pour reproduire
Edit the title of a thread by adding a dot or something. The ordering will be shuffled.
### Chamilo Version / Versión de Chamilo / Version de Chamilo
1.11.x

jderriks:
https://github.com/chamilo/chamilo-lms/issues/2125
Maybe change the default database null value to 0 also because null is never used for sticky field.

### DIFF
--- a/main/forum/forumfunction.inc.php
+++ b/main/forum/forumfunction.inc.php
@@ -3722,7 +3722,7 @@ function store_edit_post($forumInfo, $values)
         // Simple edit
         $params = [
             'thread_title' => $values['post_title'],
-            'thread_sticky' => isset($values['thread_sticky']) ? $values['thread_sticky'] : null,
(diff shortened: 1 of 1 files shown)


jmontoyaa:
Already added:

https://github.com/chamilo/chamilo-lms/commit/4175721a562f4cdef68ad43e916b49b9d9d2db9a

PR #2232, opened 2017-11-29
Update stats.php

oktavianto:
remove student can view stats.php (allow student to see the correct answer).

### DIFF
--- a/main/exercise/stats.php
+++ b/main/exercise/stats.php
@@ -3,6 +3,12 @@
 
 require_once __DIR__.'/../inc/global.inc.php';
 
+// Access control
+api_protect_course_script(true);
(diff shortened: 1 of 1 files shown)


ywarnier:
Thank you for your contribution!
I think this is incorrect (as it will remove access to the whole script). Seeing the right answer should be controlled by the settings of the exercise.
But we'll review that and come back to you.

jmontoyaa:
Good eye. Added to 1.11.x

ywarnier:
That still doesn't look like it cares about the exercise settings :-)

jmontoyaa:
Which exercise settings? That page is only to see stats about an exercise. Only available for teachers ...

ywarnier:
That answers my question then... :-)

PR #2234, opened 2017-11-30
1.11.x

aragonc:


ywarnier:
@jmontoyaa, @aragonc says he doesn't really know if that was accepted or partially integrated or what. Basically, the only conflict at the moment is with the skillsList.

PR #2286, opened 2018-01-06
Applied fixes from FlintCI

jmontoyaa:


Soullivaneuh:
The multiple PR is probably do to the manual analysis resets. Sorry about that.

PR #2289, opened 2018-01-07
Applied fixes from FlintCI

jmontoyaa:


jmontoyaa:
Yes, I agree it's an automated process using phpfixer. No harm.

PR #2474, opened 2018-04-03
 Add twig function to get template file inside template

AngelFQC:
This changes the way how template files are included or extended

Call to template with `template` variable is not useful when we are only overrides one template file

### DIFF
--- a/main/inc/lib/template.lib.php
+++ b/main/inc/lib/template.lib.php
@@ -489,12 +489,12 @@ class Template
      */
     public static function findTemplateFilePath($name)
     {
-        $sysCodePath = api_get_path(SYS_CODE_PATH);
+        $sysTemplatePath = api_get_path(SYS_TEMPLATE_PATH);
 
         // Check if the tpl file is present in the main/template/overrides/ dir
         // Overrides is a special directory meant for temporary template
         // customization. It must be taken into account before anything else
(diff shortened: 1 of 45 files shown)


ywarnier:
Could we not *also* replace the template variable (substitute it with a macro or something like that) to avoid updating the templates? (because that would also imply changing the documentation at doc.chamilo.org for the different examples, in the middle of minor versions)

AngelFQC:
I'm changing the function to a filter to make the transition less difficult

main/inc/lib/template.lib.php
```diff
@@ -479,33 +480,50 @@ public function set_header($status)
...
+     * If template not found in overrides/ or custom template folder, the default template will be used.
      *
      * @param string $name
      *
      * @return string
      */
-    public function get_template($name)
+    public static function findTemplateFilePath($name)
     {
+        $sysCodePath = api_get_path(SYS_CODE_PATH);
```
ywarnier: Please use SYS_TEMPLATE_PATH here (and modify the corresponding usage of $sysCodePath with $sysTemplatePath)

ywarnier: This is because, in the future, the template path is likely to move *out* of main/ (and into app/ or somewhere around there), and then it will be less painful to make the change if we do this change here now.

PR #2500, opened 2018-04-20
Mejoras en about session tpl

aragonc:
Mejoradas de diseño 
dis_about


main/template/default/session/about.tpl
```diff
@@ -1,275 +1,292 @@
...
-            </ul>
-        </div>
-        <div class="panel-course">
-            <div class="row">
-                {% if courses|length > 1 %}
-                    <div class="col-xs-12">
-                        <h3 class="text-uppercase">{{ course_data.course.getTitle }}</h3>
-                    </div>
-                {% endif %}
+        <div class="panel panel-default panel-course">
```
jmontoyaa: Better to use macros instead of repeat the panels everywhere  ... I think panels changed in bootstrap 4.

Issue #2378, opened 2018-02-13
Issue with LTI plugin


jacovm:
Hi

When trying to add an External Tool (LTI tool) to a course, the tool doesn't save and doesn't display anywhere.

I have enabled the IMS/LTI plugin in the admin section: 
![untitled](https://user-images.githubusercontent.com/24301344/36153212-2f7e2124-10d6-11e8-91b1-101e40ddce96.png)

In Regions I have selected all the options, and clicked "Enable the selected plugins": 
![image](https://user-images.githubusercontent.com/24301344/36153266-67c39424-10d6-11e8-9494-500ea80d1c71.png)

In the course Settings I see the IMS/LTI option 
![untitled](https://user-images.githubusercontent.com/24301344/36153367-af6ffe7a-10d6-11e8-8fe0-bc1ed1be3a44.png)

I enter all the LTI details, name , URL, key, secret. I can only select "none" under Type. Not sure if that is the cause of this issue?
![untitled](https://user-images.githubusercontent.com/24301344/36153121-e7ac2fda-10d5-11e8-9d7f-92f058b4a5f5.png)
But clicking on the + Add External Tool button doesn't seem to do anything?

And when I go back to my course, I don't see anywhere a link to the LTI tool I just tried to add?
![untitled](https://user-images.githubusercontent.com/24301344/36153509-2e5c7862-10d7-11e8-9d4d-db11271bf5b6.png)

I'm using a freshly installed version 1.11.4 which I downloaded from here https://chamilo.org/chamilo-lms/ and which seemed to already have the IMS/LTI in the plugin section.

Any idea what the issue may be?

Thanks

Jacov

NicoDucou:
Hi @jacovm 
This plugin has been under development since 1.11.4 so you should try with version 1.11.6 that has been released on 18 of january.

I just did a test on the test portal 11.chamilo.org and I got a new tool integrated in the course as you can see here https://11.chamilo.org/courses/NICODU/index.php?id_session=0 with the new tool called "Mon test".

So try to update to 1.11.6 or install 1.11.6 and test again or test it on the test portal to have an example of your problem on this test platform.

Best regards

--
Nicolas

jacovm:
Nicolas, 

Thank you for your reply. 

We upgraded our Chamilo and installed 1.11.6 so we can test.

We were able to setup the LTI tool, however it doesn't seem to launch properly. 
[...]

PR #2504, opened 2018-04-24
Fixed form.php to send to the LTI Provider using POST arguments inste…

meuhland:
Changed form.php to use POST instead of QueryString Parameters it should resolve issue #2378 and be more in line with the IMS / LTI 1.0 Spec 

>
 The Basic LTI launch protocol is a POST to the launch URL with the Basic LTI parameters described above, properly signed using OAuth.

> 
Source : https://www.imsglobal.org/specs/ltiv1p0/implementation-guide#toc-11

### DIFF
--- a/plugin/ims_lti/form.php
+++ b/plugin/ims_lti/form.php
@@ -83,8 +83,14 @@ $result = $oauth->sign(array(
         <title>title</title>
     </head>
     <body>
-        <form action="<?php echo $result['signed_url'] ?>" name="ltiLaunchForm" method="post" encType="application/x-www-form-urlencoded">
-            <input type="submit" value="Press to continue to external tool"/>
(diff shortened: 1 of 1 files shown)


ywarnier:
@AngelFQC puedes revisar este PR (para ponerlo en la 1.11.x) o necesitas algún ejemplo LTI?

ywarnier:
This has been merged into 1.11.x and master. Thank you @meuhland for this contribution and sorry for the late integration.

jacovm:
@ywarnier, I just installed a fresh Chamilo instance to test the LTI. I can enable LTI in the LMS plugins - so far so good.

However, when I try to enable LTI in Course Settings and select "Add External Tool", I get an error saying:
"You are not allowed to see this page. Either your connection has expired or you are trying to access a page for which you do not have the sufficient privileges." 

Please see screenshots below:

untitled


untitled


I get the same error in Edge, Chrome and IE. 

Any ideas what may be wrong?

Thank you

Jacov

PR #2542, opened 2018-06-01
Check the logic in method setAnswer on Ims2Question

fitorec:
The types `MCUA`, `MCMA` and `TF` generate the same answer(`ImsAnswerMultipleChoice`).

The `default` option was eliminated, since it was inefficient besides generating the same result, see the simple test:

```php
function test1($type) {
	switch ($type) {
		case 1:
		case 2:
			$answer = "1 or 2";
			return $answer;
		default:
			$answer = null;
			break;
	}
	return $answer;
}

var_dump(test1(1)) . "\n";
// string(6) "1 or 2"
var_dump(test1(5)) . "\n";
// NULL

function test2($type) {
	switch ($type) {
		case 1:
		case 2:
			$answer = "1 or 2";
			return $answer;
	}
	return null;
}

var_dump(test2(1)) . "\n";
// string(6) "1 or 2"
var_dump(test2(5)) . "\n";
// NULL
```

if in structure switch in each case returns the $answer, I think it could look like:

```php
public function setAnswer()
{
	switch ($this->type) {
		case MCUA:
		case MCMA:
		case TF:
			return new ImsAnswerMultipleChoice($this->id);
		case FIB:
			return new ImsAnswerFillInBlanks($this->id);
		case MATCHING:
		case MATCHING_DRAGGABLE:
			return new ImsAnswerMatching($this->id);
		case FREE_ANSWER:
			return new ImsAnswerFree($this->id);
		case HOT_SPOT:
			return new ImsAnswerHotspot($this->id);
	}
	return null;
}
```

### DIFF
--- a/main/exercise/export/qti2/qti2_classes.php
+++ b/main/exercise/export/qti2/qti2_classes.php
@@ -19,13 +19,7 @@ class Ims2Question extends Question
     {
         switch ($this->type) {
             case MCUA:
-                $answer = new ImsAnswerMultipleChoice($this->id);
-
-                return $answer;
             case MCMA:
-                $answer = new ImsAnswerMultipleChoice($this->id);
-
-                return $answer;
             case TF:
                 $answer = new ImsAnswerMultipleChoice($this->id);
 
@@ -47,12 +41,9 @@ class Ims2Question extends Question
                 $answer = new ImsAnswerHotspot($this->id);
 
                 return $answer;
-            default:
-                $answer = null;
-                break;
         }
 
-        return $answer;
+        return null;
     }
 
     public function createAnswersForm($form)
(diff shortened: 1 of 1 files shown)


ywarnier:
@fitorec I was going to say that our coding conventions force you to have a default in a switch, but that is not the case. However, it is important to always have a default value anyway, because the code of Chamilo is ever changing and that we need to account for un-planned values that would be added in another part of the code.

fitorec:
I think that the variable `$answer` should be defined outside of the `switch` structure

```php
        $answer = null;
        switch ($this->type) {
            case MCUA:
            case MCMA:
            case TF:
                $answer = new ImsAnswerMultipleChoice($this->id);

                return $answer;
            // . . .
            default:
                break;
        }
        return $answer;
```

What is your opinion?

Issue #2568, opened 2018-06-18
EUGDPR: Add data display + export screen (art.20)


ywarnier:
Article 20 requires the possibility for any user to require a structured format of all the personal data the organization keeps about the user.

We should have an export button on the data info page that generates a JSON file with all that data.

The data info page should contain the data specified in #2503

jmontoyaa:
This is not the same as #2606 ?

NicoDucou:
#2606 includes this task.
But from what I can see of the page on main/social/personal_data.php there are still some information missing as referenced to the list in #2503

ywarnier:
Yes, it is a bit confusing now, I admit that. #2606 is the PR ID. It should have been associated with this issue here, so it's kind of a duplicate that will disappear when we merge (which I will do as soon as I have checked your changes).

ywarnier:
There you go, branch 2568 merged and #2606 closed.
We can continue for the missing info from #2503 here.

ywarnier:
Now when I go to the personal_data.php page, I get this circular reference error:
![serializer-error](https://user-images.githubusercontent.com/609839/43536177-907c329c-9581-11e8-90e1-16dfc468bc21.png)

Fatal error: Uncaught Symfony\Component\Serializer\Exception\CircularReferenceException: A circular reference has been detected when serializing the object of class "Chamilo\UserBundle\Entity\User" (configured limit: 1) in /var/www/chamilo111x/vendor/symfony/serializer/Normalizer/AbstractNormalizer.php on line 226
Symfony\Component\Serializer\Exception\CircularReferenceException: A circular reference has been detected when serializing the object of class "Chamilo\UserBundle\Entity\User" (configured limit: 1) in /var/www/chamilo111x/vendor/symfony/serializer/Normalizer/AbstractNormalizer.php on line 226

jmontoyaa:
I don't have that error. I added a fix just in case.

ywarnier:
Now it works for me.
Now we need to fill the data with the related collections (call foreach on each relation, as you suggested?) for #2503

jmontoyaa:
Which collections? All of them? I add it the extra fields here:

https://github.com/chamilo/chamilo-lms/commit/a7567be9c54fbd1ed16779479b7a051c76c3a282

NicoDucou:
>Which collections? All of them?
@jmontoyaa in task #2503 I've listed all the personal data related to a user.
We need to present all of them on this page.
[...]

PR #2588, opened 2018-07-16
WIP - #2568

ywarnier:
Work in progress with the private data treatment information


main/admin/legal_list.php
```diff
@@ -8,6 +8,8 @@
  */
 $cidReset = true;
 require_once __DIR__.'/../inc/global.inc.php';
+use Chamilo\CoreBundle\Entity\Repository\LegalRepository;
```
jmontoyaa: The "use" should be in the top before global.inc check other files.

main/social/personal_data.php
```diff
@@ -0,0 +1,177 @@
...
+api_block_anonymous_users();
+if (!api_get_configuration_value('enable_gdpr')) {
+    api_not_allowed(true);
+}
+
+$userId = api_get_user_id();
+
+$entityManager = Database::getManager();
+/** @var UserRepository $repository */
+$repository = $entityManager->getRepository('ChamiloUserBundle:User');
```
jmontoyaa: You can use "api_get_user_entity($userId)", There are also "api_get_course_entity()" and "api_get_session_entity() "functions.

main/social/personal_data.php
```diff
@@ -0,0 +1,177 @@
...
+$icon = Display::return_icon('export_excel.png', get_lang('Export'), null,ICON_SIZE_MEDIUM);
+$personalData['data_export_icon'] = $icon;
+$personalData['permissions'] = $termsAndConditionsAcceptance;
+$personalData['responsible'] = api_get_setting('personal_data_responsible_org');
+
+$em = Database::getManager();
+/** @var LegalRepository $legalTerms */
+$legalTermsRepo = $em->getRepository('ChamiloCoreBundle:Legal');
+//Get data about the treatment of data
+$treatmentTypes = [
```
jmontoyaa: If the types (101, 102, etc) are hardcode values they should be define in some library legal.lib, or better in the entity ...

src/Chamilo/CoreBundle/Entity/TrackELogin.php
```diff
@@ -18,9 +18,14 @@ class TrackELogin
     /**
      * @var int
```
jmontoyaa: Just as a reminder, this "@ orm" changes mean changes in the database. So they are not allowed in 1.11.x.
You can remove the "@" and then doctrine will ignore those changes, then in configuration.dist you can add a note to add the "@".

ywarnier:
@jmontoyaa I pushed this to a branch of this project directly. Could you clean up for the two comments you mentioned and for which I have some doubts still? (won't have time to do that properly and I want this to move forward quickly and there's still some work). Work on the branch with me for now, then we'll merge in a few days.

ywarnier:
Moved to #2606

Issue #2687, opened 2018-09-27
Course About Page shows Subscription button even if Course doesn't allow to be subscribed to


meuhland:
### Current behavior / Resultado actual / Résultat actuel
About : 
![image](https://user-images.githubusercontent.com/6123610/46166145-7a577100-c260-11e8-88f3-0cc592f360f4.png)

Course Setting : 
![image](https://user-images.githubusercontent.com/6123610/46166129-67dd3780-c260-11e8-8768-0c3032daa338.png)

Result : 
![image](https://user-images.githubusercontent.com/6123610/46166182-90fdc800-c260-11e8-9f47-931d26e2d3c2.png)

### Expected behavior / Resultado esperado / Résultat attendu

Not being allowed to subscribe as per course setting in the first place.

### Steps to reproduce / Pasos para reproducir / Étapes pour reproduire

Create a public course that doesn't allow subscription by the user, and as a non-admin try to enrol via the about page of the course.

### Chamilo Version / Versión de Chamilo / Version de Chamilo
1.11.8

ywarnier:
Hi @meuhland 
Could you describe your use case? Is that useful in some cases to have a public course to which you cannot subscribe? They can still access the course even without subscribing, right?
It's just to be able to judge the priority...

meuhland:
That’s exactly my point as of right now you can set a public course that
only administrator can subscribe to but the button will still be visible in
the about page of the course (via the course catalog).
Which results in a redirection to the class with an error saying something
generic like unknown error contact the portal admin.

More or less it boils down to adding the logic to check if you can
subscribe to the course then display the subscribe button in the about page.

meuhland:
It’s a 6 or 7 line fix if you just hide it when not allowed I can PR it
tomorrow (got it running on our internal staging server)

ywarnier:
I was going to suggest to help you out for the patch (not because we don't want to do it but because we'd like to help our community to be more participative). Will gladly review and merge your PR.

meuhland:
You know the organization I work for and know of our plans going forwards
with Chamilo. So you know we’re working a common goal here. :)

ywarnier:
True... I always forget :-)

PR #2688, opened 2018-09-28
A logic condition as described in Issue #2687 was showing subscribe when it shouldn't

meuhland:
Fixed the logic by reusing condition that was causing the unknown error.

Enjoy!

### DIFF
--- a/main/course_info/about.php
+++ b/main/course_info/about.php
@@ -124,7 +124,16 @@ $topics = [
 ];
 
 $subscriptionUser = CourseManager::is_user_subscribed_in_course($userId, $course->getCode());
-
+$course_table = Database::get_main_table(TABLE_MAIN_COURSE);
+$sql = "SELECT code, visibility FROM $course_table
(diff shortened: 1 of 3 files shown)


meuhland:
That's what you get for using vi to fix code right on the server...

meuhland:
I'll repost without the garbage from me hitting the wrong button

main/course_info/about.php
```diff
@@ -124,7 +124,16 @@
 ];
 
 $subscriptionUser = CourseManager::is_user_subscribed_in_course($userId, $course->getCode());
-
+$course_table = Database::get_main_table(TABLE_MAIN_COURSE);
```
jmontoyaa: You can use:

`$courseInfo = api_get_course_info_by_id($courseId);`

Then later you can check the subscribe value:

`$courseInfo['subscribe']`

meuhland: Awesome then it's even simpler to fix if this return a boolean.

I'd just be a assign and the same change in the template.

PR #2690, opened 2018-09-28
A logic condition as described in Issue #2687 was showing subscribe when it shouldn't in course about page

meuhland:
Fixed the logic by reusing condition that was causing the unknown error. 

Enjoy!


meuhland:
Will fix it according to @jmontoyaa comment on the closed one.

meuhland:
@jmontoyaa  There you can review this one it includes the feedback from earlier.

main/course_info/about.php
```diff
@@ -165,6 +172,7 @@
 $essence = Essence\Essence::instance();
 $template->assign('essence', $essence);
 $template->assign('is_premium', $courseIsPremium);
+$template->assign('allowSubscribe', $allowSubscribe);
```
jmontoyaa: Should be 'allow_subscribe'

then in the template you can call it {% if allow_subscribe == true %}

main/course_info/about.php
```diff
@@ -125,6 +125,13 @@
 
 $subscriptionUser = CourseManager::is_user_subscribed_in_course($userId, $course->getCode());
 
+$courseInfo = api_get_course_info_by_id($courseId);
+$allowSubscribe = null;
+if ($courseInfo['subscribe'] == true || api_is_platform_admin()) {
```
jmontoyaa: You can first set it false so you can avoid the else

```
$allowSubscribe = false;
if ($courseInfo['subscribe'] == true || api_is_platform_admin()) {
      $allowSubscribe = true;
}
```

meuhland:
You are completely right... I just hacked that whole thing quickly
yesterday to close a bug here internally will fix in a minute

main/course_info/about.php
meuhland: Fixed!

PR #2715, opened 2018-10-30
Changes in translation to pt-br lang

HelenDias:


ywarnier:
Hi @HelenDias and thanks for your attempted PR. You might want to give us a little bit of detail on what you are trying to achieve, because for Chamilo 1.11.*, you should use https://translate.chamilo.org, and for the master branch (future 2.0) our translation system is not ready yet.

PR #2744, opened 2018-12-01
Update email_editor.php

hichamon306:
fix email send with user  address

### DIFF
--- a/main/inc/email_editor.php
+++ b/main/inc/email_editor.php
@@ -77,8 +77,7 @@ if ($form->validate()) {
             $email_administrator,
             $title,
             $text,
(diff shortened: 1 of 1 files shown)


jmontoyaa:
I think he's right, when sending an email, it should be using the default sender (portal admin email), instead of using the current user email, because sometimes there are some problems using another platform email ... But at the same time I didn't create that page so I'm not sure what's the purpose of it ... :)

jmontoyaa:
Ok, @ywarnier this change should be in 1.11.x then copied to master? Or only in master?

ywarnier:
@hichamon306 This is included in https://github.com/chamilo/chamilo-lms/pull/2789/files now, so I'm just closing this one. Thanks!

PR #2778, opened 2019-01-16
Xml entity fixes

AngelFQC:


AngelFQC:
WIP

QTI imports xml, but is hard replace the current method to parse xml in qti import with Symfony's Crawler

AngelFQC:
Import from QTI and Scorm done.

ywarnier:
Impressive.

PR #2959, opened 2019-07-23
1.11.10 Course Import User

fabiowoj:
The new version of chamilo, gets error on import csv to subscribe users on file main/admin/course_user_import.php

To fix the problem had to change line 109
from
`$result = CourseManager::subscribeUser(`

to
`$result = CourseManager::subscribe_user(`


jmontoyaa:
I closed it because there's a mix of branches between 1.11.x and master.

PR #3015, opened 2019-10-30
make unsubcribe message unambiguous

dranjor:
the old text "Abmelden" can be understood as "logout"


ywarnier:
Same here as in #3016

ywarnier:
What I said in #3016 is now deprecated: we will import the language files again before version 2.0, so there's nothing more to do here.

PR #3016, opened 2019-10-30
fix german translation - give password reset url

dranjor:


ywarnier:
I've sent your update but in a different way. We don't manage translations through Git, but through a separate system. We will have a bit of an issue though, because we are currently developing in the master branch for a release in early 2020, and this doesn't work with the same language translation mechanism, so you will have to resubmit this to the master branch once the German language has been enabled. I will mark this PR as unresolved and to be planned for version 2.0, but just so you know: it has been sent to the 1.11.x branch and will appear in Chamilo 1.11.12. Thanks!

Issue #3022, opened 2019-11-08
Composer json file sorting


marvil07:
**Is your feature request related to a problem?**
Dependencies are easier to find if sorted alphabetically.

**Describe the solution you'd like**
Use alphabetical order for all things inside composer.json.

**Additional context**
When merging and when searching dependencies or other things in the file, it is easier if things are sorted.

marvil07:
@ywarnier Thanks for reviewing and merging #3023. 
Now I opened an analogous PR #3064 for `master` (`2.x`).

ywarnier:
This has been merged. Thanks for the PR.

PR #3023, opened 2019-11-08
Sort things alphabetically on composer.json.

marvil07:
See #3022


ywarnier:
Not sure if the issue detected by Travis CI is correct, but launching again just to make sure (I don't want some library not to be installed or something like that because of a weird dependency manager issue)

composer.json
```diff
@@ -5,118 +5,115 @@
...
+        "graphp/graphviz": "~0.2.0",
+        "guzzlehttp/guzzle": "~6.0",
         "imagine/imagine": "0.6.3",
+        "ircmaxell/password-compat": "~1.0.4",
+        "jbroadway/urlify": "1.1.0-stable",
         "jeroendesloovere/vcard": "~1.7",
-        "knplabs/doctrine-behaviors": "~1.1",
         "jimmiw/php-time-ago": "0.4.15",
-        "phpoffice/phpword": "~0.14",
-        "facebook/php-sdk-v4": "~5.0",
```
ywarnier: @marvil07 This one is missing in the new version. I know it says it's deprecated, but removing it in a sorting commit is not a good idea (and yes, I'm checking those one by one because it's very important)

composer.json
```diff
@@ -5,118 +5,115 @@
...
-        "sylius/attribute": "0.13.0",
-        "sylius/translation": "0.13.0",
-        "patchwork/utf8": "~1.2",
-        "ddeboer/data-import": "@stable",
-        "league/csv": "~8.0",
-        "phpoffice/phpexcel": "~1.8",
-        "webit/eval-math": "1.0.1",
-        "clue/graph": "~0.9.0",
-        "graphp/graphviz": "~0.2.0",
+        "facebook/php-sdk-v4": "~5.0",
```
marvil07: Here it was.

composer.json
marvil07: @ywarnier Please see my comment, I pointed where in the diff it is. I may be missing something here.

marvil07: I meant https://github.com/chamilo/chamilo-lms/pull/3023#discussion_r345497297

composer.json
ywarnier: weird, my mistake.

marvil07:
:+1:

PR #3059, opened 2019-12-24
Minor - lang variables

nosolored:
Insertion of Basque translation strings


ywarnier:
This was made manually to allow it to go through translate.chamilo.org
(see 3391cb14da)

PR #3136, opened 2020-03-26
update translation.

spacecabbie:
Update translation got more coming translate.chamilo.org seems to be in transition ?


ywarnier:
Hi @spacecabbie 

Sorry for the delay, translation.chamilo.org is still the right platform to send translations at this time. Do you have a working account on it?

spacecabbie:
I have an account but am still unable to edit or add translations on the platform i have emailed @ywarnier about it.

Issue #3096, opened 2020-02-20
Administration page blocks navigation for a while when lacking direct internet access


sebastiendu:
**Description**
On Chamilo instances installed on a server that does not have direct access to remote resources such as https://version.chamilo.org/news/latest.php, the administration page appears but blocks any subsequent navigation for a while.

**To Reproduce**
1. Install Chamilo on a server behind a firewall that block outgoing http requests.
2. Login as administrator and open the Administration page (/main/admin/index.php)
3. Don't wait and click on "User list", "Session list" or any other link.

**Expected behavior**
User list should appear within a reasonable delay.

**Server**
 - Chamilo 1.11.x of 2020-02-04
 - no access to remote https locations

**Additional context**
Setting $_configuration['proxy_settings'] (on servers that have an HTTP proxy available) does not seem to help.

sebastiendu:
(not verified hypothesis)

The HTML page generated by /main/admin/index.php and admin/settings_index.tpl
contains a call to /main/inc/ajax/admin.ajax.php?a=get_latest_news

admin.ajax.php first loads /main/inc/global.inc.php, which indirectly calls start_session();
then admin.ajax.php calls function getLatestNews() without first closing the PHP open session.
While a PHP session is open, it is locked. Through start_session(), other pages wait for the lock to be released to proceed.

https://github.com/chamilo/chamilo-lms/blob/85c425f910f42f16108f001dead3038eb14caf0e/main/inc/ajax/admin.ajax.php#L244

sebastiendu:
3 years ago, the same issue was fixed for
/main/inc/ajax/admin.ajax.php?a=version

https://github.com/chamilo/chamilo-lms/commit/c5120d60dbc60432580da375ccc70a484a965328

ywarnier:
Fixed by PR

PR #3138, opened 2020-03-26
Set relevant environment variables http*_proxy - refs BT#16686

### LINKED ISSUE #3117: $_configuration['proxy_settings'] is not used everywhere
**Describe**
Chamilo attempts to send HTTPS requests without using the HTTPS proxy configuration when available.

**To Reproduce**

0. Install Chamilo on a server with no direct internet access
1. Go to /main/admin
2. See the error message appear in red under Version Verification
3. Define $_configuration['proxy_settings'] in app/config/configuration.php
4. Reload /main/admin
5. See the error message appear in red under Version Verification

**Expected behavior**
No error message under Version Verification

**Server**
 - Version of Chamilo: 1.11.x of 2020-02-04
 - Version of PHP: 7.2

**Additional context**
We should scan source code for any use of Curl, Guzzle and / or others so to detect all forgotten use of parameter 'proxy_settings'.

sebastiendu:
fixes #3117 and in some cases #3096

### DIFF
--- a/main/inc/global.inc.php
+++ b/main/inc/global.inc.php
@@ -82,11 +82,14 @@ api_request_uri();
 
 // Set web proxy environment variables
 foreach ([
-             api_get_configuration_sub_value('proxy_settings/stream_context_create/https/proxy'),
-             api_get_configuration_sub_value('proxy_settings/stream_context_create/http/proxy'),
-             api_get_configuration_sub_value('proxy_settings/curl_setopt_array/CURLOPT_PROXY'),
-         ] as $value) {
-    if ($value && is_string($value)) {
+             'proxy_settings/stream_context_create/https/proxy',
+             'proxy_settings/stream_context_create/http/proxy',
+             'proxy_settings/curl_setopt_array/CURLOPT_PROXY',
+         ] as $path) {
+    $value = api_get_configuration_sub_value($path);
+    if (!empty($value) && is_string($value)) {
+        // libcurl reads environment variable https_proxy: https://curl.haxx.se/libcurl/c/libcurl-env.html
+        // \GuzzleHttp\Client::configureDefaults reads environment variable HTTPS_PROXY
         foreach (['https_proxy', 'http_proxy', 'HTTPS_PROXY', 'HTTP_PROXY'] as $envVar) {
             if (false === getenv($envVar)) {
                 putenv("$envVar=$value");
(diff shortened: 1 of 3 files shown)


main/inc/global.inc.php
```diff
@@ -80,6 +80,22 @@
 // Fix bug in IIS that doesn't fill the $_SERVER['REQUEST_URI'].
 api_request_uri();
 
+// Set web proxy environment variables
+foreach ([
+             api_get_configuration_sub_value('proxy_settings/stream_context_create/https/proxy'),
+             api_get_configuration_sub_value('proxy_settings/stream_context_create/http/proxy'),
+             api_get_configuration_sub_value('proxy_settings/curl_setopt_array/CURLOPT_PROXY'),
+         ] as $value) {
+    if ($value && is_string($value)) {
```
ywarnier: I would prefer it if you used (!empty($value) && is_string($value)) here, because historically, PHP has not been very good at dealing with undefined variables properly (so a warning or notice could appear here depending on the PHP version and configuration.

main/inc/global.inc.php
```diff
@@ -80,6 +80,22 @@
...
+        foreach (['https_proxy', 'http_proxy', 'HTTPS_PROXY', 'HTTP_PROXY'] as $envVar) {
```
ywarnier: I would add this comment here (it is in your bug report but there's no clear reason explained here as to why we use lowercase and uppercase versions of the same):
// libcurl reads environment variable https_proxy: https://curl.haxx.se/libcurl/c/libcurl-env.html
// Guzzle reads environment variable HTTPS_PROXY

main/inc/global.inc.php
```diff
@@ -80,6 +80,22 @@
...
+            if (false === getenv($envVar)) {
+                putenv("$envVar=$value");
```
ywarnier: Could you make sure that CURL and Guzzle use getenv() and not $_ENV?
Because of this comment: https://www.php.net/manual/en/function.putenv.php#98540, I'm a bit worried that it could not get what you put in ENV through putenv().

main/inc/lib/api.lib.php
```diff
@@ -8653,6 +8653,42 @@ function api_get_configuration_value($variable)
     return false;
 }
 
+/**
+ * Retreives and return a value in an array tree configuration item
```
ywarnier: returns

ywarnier: in a hierarchical configuration array ?

main/inc/global.inc.php
sebastiendu: \GuzzleHttp\Client::configureDefaults uses getenv().
Curl reads environment variables directly (coded in C).
[To set real environment variables, you must use putenv()](https://www.php.net/manual/en/reserved.variables.environment.php#114811)

PR #3170, opened 2020-04-07
CAS Login Button form action is REQUEST_URI - refs BT#16686

sebastiendu:


ywarnier:
Looks good to me. However, we will have to watch existing CAS users in case they report any issue... To my knowledge, there are only a few. Pinging @jmontoyaa so he knows about this change too.

Issue #3172, opened 2020-04-08
tests/random questions Randomization can make test unusable


mk111:
**Describe**
when changing parameter nr. of random questions of any type of randomized test, the test can become unusable in the meaning of intend usability: test originally defined containing 10 questions can be unintendedly transformed to test containing only 2 questions . picture is attached.
btw: there is no workaround with the means of portal; changes of the test type/parameters are of no use

**To Reproduce**
1. Define new test with min. 4 question
2. set mode of the test to Random (don't use cat.), set 2 for nr. of quest., save
3. same test: set mode of the test to Random (don't use cat.), set 3 for nr. of quest., save
4. same test: set mode of the test to Random (don't use cat.), set 3 for nr. of quest., save
5. edit test, see list of the questions (2 questions should be listed)

tracked changes on c_quiz_rel_questions on column _order while changing nr. of questions:
```
original    -->     (2)   -->      (3)   -->   (1)
1                         0                 0             0
2                         1                 1             1
3                         3                 2             0
4                         4                 0             0
```
so, look at the number of different values in the last column

**Expected behavior**
to see

**Screenshots**
![randomized](https://user-images.githubusercontent.com/4478116/78774428-f32aa380-7994-11ea-95c5-6ff4c6f59ac8.png)

**Additional context**
applies to  1.11.10, i do not remember to be seen on 1.11.8 at least not been noticed 
possible related to #3047

mk111:
I can see, this one is present in current 11.x also

question: is it order of the question in the quiz/test still important when we use Randomization with any kind of question number limitation ? current implementation do not care about it.

i do not have type to test this 'fix' properly, sorry

ywarnier:
> I can see, this one is present in current 11.x also

I can't. I tried your procedure with a 10 questions test and I *always* got the number of questions I had chosen.

> question: is it order of the question in the quiz/test still important when we use Randomization with any kind of question number limitation ? current implementation do not care about it.

No, it isn't important. This c_quiz_rel_question.question_order field is only used when not random or when using random categories with ordered questions. The real order of the questions for each user is stored in the track_e_exercises table (as a comma-separated string) and is generated just when the user enters the exam (after that it is static for the rest of the exam so that an interrupted exam can be retaken later on).

> proposed fix #3173

Although I don't see it doing any harm, I haven't seen it fix anything either... For me it results in exactly the same question_order, whether I use the PR or not...

dauti:
[...]

PR #3173, opened 2020-04-08
possible fix for issue #3172

mk111:
provides all the questions for visualization of  randomized test ; keep random selected order in the way as before it was before change

putting this to 1.11.10-releases, maybeeeee somebody want to use this


ywarnier:
As stated in #3172, I cannot reproduce the issue and this PR doesn't change anything for me.

PR #3174, opened 2020-04-08
Hide buttons for admin level actions on reports tab view

juan-cortizas-ponte:
The reporting tab is displaying the action buttons to add "Followed users" and "Followed sessions" to users with the teacher role and non admin permissions.

reporting

For non admin users both actions would result in an error.

error

The proposal is to hide these two buttons assigning the $is_platform_admin variable on [/main/mySpace/index.php](https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/mySpace/index.php) to the template

fix_01

and checking if the template ([main/template/default/my_space/index.tpl](https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/template/default/my_space/index.tpl)) has to draw the buttons for both actions.

fix_02

With this fix the buttons aren´t drawed to users with teacher role and without admin permissions

reporting_fixed

### DIFF
--- a/main/mySpace/index.php
+++ b/main/mySpace/index.php
@@ -298,6 +298,7 @@ $view->assign(
 $view->assign('studentboss', STUDENT_BOSS);
 $view->assign('drh', DRH);
 $view->assign('stats', $stats);
+$view->assign('isPlatformAdmin', $is_platform_admin);
 
 $form = new FormValidator(
     'search_user',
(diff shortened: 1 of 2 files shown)


jmontoyaa:
"isPlatformAdmin" is not needed, we have a variable "_u.is_admin", I accepted this PR and later I applied the change here: 

https://github.com/chamilo/chamilo-lms/commit/8d1c20913ab3189645f16a8862f7a466aa1642ab

It should be the same.

Issue #3171, opened 2020-04-08
tests/question pool : when question is reused it appears duplicated (tripled, ..


mk111:
**Describe**
as seed

**To Reproduce**
1. create exercise
2. add some questions
3. create another exercise
4. add questions from question pool
5. show pool

**Expected behavior**
single line for single question (data about how many times quest. is used can be of use)

**Additional context**
adding following to line 781 of main/exercises/question_pool.php is good enough for temporary fix:
`$aabbcc=[]; foreach($mainQuestionList as $q001) $aabbcc[$q001['id']]=$q001; $mainQuestionList=$aabbcc; unset($aabbcc);`
(showing exercise_id in query for constructing $mainQuestonList preventing distinc to be selective on questions)

jmontoyaa:
This was already fixed already.

PR #3176, opened 2020-04-10
Question reused appears duplicated #3171

mk111:
looks ok, there was no complaint so far


ywarnier:
As stated in #3171, this was "already fixed already" ;-)
Thanks for your contribution anyway! It could have helped otherwise.

Issue #3177, opened 2020-04-10
Tests/Report by question broken for many types of tests


mk111:
**Describe**
There is a bug in report whenever you select any type of test where the option to choose M questions out of N is present on execution.  After execution, just  N questions is presented in report!
Than, not the same N question is shown every time; as question selection thr report is pseudo random to.  This report should  contain all possible questions that are possible in the test, so be the same as 
there is no question draw.

**To Reproduce**
1. define 4 questions of any question type
2. put them into different 2 categories, by 2 per cat.
3. execute test 3 or 4 time
4. view report by question, put down or remember list of the questions
5. refresh browser, compare with previous report
if no change:
6. refresh browser again ,...

**Expected behavior**
this report should  contain all possible questions from the test **regardless** of  draw.

**Screenshots**
outer image shows report done with no draw(as should be), inner (inside rect.) current report

![rep_by_question_good](https://user-images.githubusercontent.com/4478116/79004025-2f583280-7b54-11ea-8c84-76f22e7cceb2.png)

**Additional context**
applies to 1.11.x, , 1.11.10

ywarnier:
The latest fixes should solve this issue on https://11.chamilo.org.
@mk111 please check when you get the opportunity.

ywarnier:
Assigned to @NicoDucou for review

NicoDucou:
I just reproduced the case with this test : https://11.chamilo.org/main/exercise/overview.php?cidReq=GITHUB3177&id_session=0&gidReq=0&gradebook=0&origin=&exerciseId=451
I realised the test 4 times and I had different questions based on the 2 categories which showed 1 question per category.

On the report per question page I can see all the questions : https://11.chamilo.org/main/exercise/stats.php?cidReq=GITHUB3177&id_session=0&gidReq=0&gradebook=0&origin=&exerciseId=451

So the problem is fixed for this.

PR #3178, opened 2020-04-10
Corrects report by question stat ref. #3177

mk111:
did not see this file used anywhere else or for anything else but i'm a bit worried about so much checkings on empty list of the questions, but cant' see the case this can happened (before the change and now)


mk111:
couses another issue

PR #3261, opened 2020-05-21
If the BigBlueButton URL has a protocol, use it when making API calls

daronco:
When configuring a BigBlueButton server with HTTPS, it wasn't using HTTPS in the API calls. Even if the server returned a redirect on HTTP calls they would not work, since curl is not configured to follow redirects. So all API calls would fail.

This was initially done and tested over the tag v1.11.10, then ported to branch 1.11.x.

I've seen that you recently added the `CURLOPT_FOLLOWLOCATION` flag (https://github.com/chamilo/chamilo-lms/commit/79b4840172b3f99bc41b603eac30167f5cd45c7c), so I would probably not run into this issue in a new version. However, I still think that using the protocol the user specified and not having to follow redirects is a better solution.

### DIFF
--- a/plugin/bbb/lib/bbb.lib.php
+++ b/plugin/bbb/lib/bbb.lib.php
@@ -125,13 +125,15 @@ class bbb
                 $urlWithProtocol = $bbb_host;
             } else {
                 // We assume it's an http, if user wants to use https, the host *must* include the protocol.
-                $urlWithProtocol = 'http://'.$bbb_host;
+                $this->protocol = 'http://';
+                $urlWithProtocol = $this->protocol.$bbb_host;
             }
 
             // Setting BBB api
             define('CONFIG_SECURITY_SALT', $this->salt);
             define('CONFIG_SERVER_URL_WITH_PROTOCOL', $urlWithProtocol);
             define('CONFIG_SERVER_BASE_URL', $this->url);
+            define('CONFIG_SERVER_PROTOCOL', $this->protocol);
 
             $this->api = new BigBlueButtonBN();
             $this->pluginEnabled = true;
--- a/plugin/bbb/lib/bbb_api.php
+++ b/plugin/bbb/lib/bbb_api.php
@@ -41,6 +41,7 @@ class BigBlueButtonBN
 {
 	private $_securitySalt;
 	private $_bbbServerBaseUrl;
+	private $_bbbServerProtocol;
 
 	public function __construct()
     {
@@ -51,6 +52,7 @@ class BigBlueButtonBN
 		// simply flow in here via the constants:
 		$this->_securitySalt 		= CONFIG_SECURITY_SALT;
 		$this->_bbbServerBaseUrl 	= CONFIG_SERVER_BASE_URL;
+		$this->_bbbServerProtocol 	= CONFIG_SERVER_PROTOCOL;
 	}
 
 	private function _processXmlResponse($url)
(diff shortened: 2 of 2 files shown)


AngelFQC:
BBB works with http?

lonesomewalker:
> Does BBB works with http?

Kind of.
Yes, you can do it http-based, but you won't get up a real meeting running, since WebRTC uses https.

Vote for reject.

daronco:
It's not a matter of whether BigBlueButton works or not with HTTP. If you configure this plugin to use HTTPS right now, it's removing the protocol and making API calls without it, so they will always be using HTTP. So, unless the server is serving the API using HTTP, the API calls fail (because 1.11.10 doesn't have `CURLOPT_FOLLOWLOCATION` set to follow redirects.

Or am I doing something wrong?

HTTPS is recommended and it's the default in BigBlueButton. But the API should work with HTTP if, for any reason, you would need it to.

In any case, there are custom BigBlueButton deployments, there are load balancers, and there are other applications that implement BigBlueButton's API. I believe it's important for integrations to support HTTPS and HTTP as well.

lonesomewalker:
> In any case, there are custom BigBlueButton deployments, there are load balancers, and there are other applications that implement BigBlueButton's API. I believe it's important for integrations to support HTTPS and HTTP as well.

If it is not working, why support a non-functional setup?
If you go for load balancers, also Scalelite recommends http, and the connection is done by https to the endpoints.

So, i kindly ask you to better fix your broken setups instead of bloating other projects ;-)
In the end, it needs more support than a stable, functional version.
Also, there might occur other side effects for Chamilo (not tested so far, but who knows...).

daronco:
Seems like the issue is not clear yet, so I'll add some more info. The reason this integration with BigBlueButton works is probably only because the standard BigBlueButton installation serves the API using both HTTP and HTTPS. All API calls going from Chamilo to BigBlueButton are using HTTP, even if you configure an URL with HTTPS in Chamilo.

As I pointed out in the initial message, you even have a "fix" for Scalelite in [this commit](https://github.com/chamilo/chamilo-lms/commit/79b4840172b3f99bc41b603eac30167f5cd45c7c). Probably Scalite redirects from HTTP to HTTPS, so you had to turn on the flag to follow redirects.

I'm proposing a different solution: to use the protocol the user configured in the URL instead of relying on redirects. It's faster and more secure (since you're actually using HTTPS for API calls).

I don't believe that less than 10 lines of code for enabling HTTPS in API calls from Chamilo to BigBlueButton can be considered "bloating other projects".

lonesomewalker:
> All API calls going from Chamilo to BigBlueButton are using HTTP, even if you configure an URL with HTTPS in Chamilo.

OUCH, just stepped through your pull request.
This really looks nasty!

> As I pointed out in the initial message, you even have a "fix" for Scalelite in [this commit](https://github.com/chamilo/chamilo-lms/commit/79b4840172b3f99bc41b603eac30167f5cd45c7c). Probably Scalite redirects from HTTP to HTTPS, so you had to turn on the flag to follow redirects.

No, since you can run Scalelite without https, only the endpoints will be https.
So in general you are right.

> I'm proposing a different solution: to use the protocol the user configured in the URL instead of relying on redirects. It's faster and more secure (since you're actually using HTTPS for API calls).

How can this be more secure?
Wouldn't it be better to only allow https, because otherwise the trouble begins?

> I don't believe that less than 10 lines of code for enabling HTTPS in API calls from Chamilo to BigBlueButton can be considered "bloating other projects".

Don't get me wrong: i wrote the comments before without having had a look at your pull request (shame on me).

ywarnier:
@daronco Thanks for spotting this. Yes, contacting in HTTP to later get a redirect to HTTPS is insecure.

I suppose we could say that if the user hasn't given any protocol, there is a better chance of the call working on HTTP (default configuration for any web server as you can't auto-generate a certificate for each and every host you declare on a server), otherwise I would have gone a step further and used HTTPS whenever the user didn't declare any protocol.

Thank you for your contribution! Most appreciated.

PR #3284, opened 2020-06-07
Add dutch lang

spacecabbie:
quick addition for dutch

### DIFF
--- a/plugin/static/lang/dutch.php
+++ b/plugin/static/lang/dutch.php
@@ -6,7 +6,7 @@
  * @translation Hans Haufe <hans@cappon.org>
  */
 $strings['plugin_title'] = "Statisch";
-$strings['plugin_comment'] = "Statische html inhoud weergeven.;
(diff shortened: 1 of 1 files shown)


plugin/static/lang/dutch.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
+ * @copyright (c) 2012 University of Geneva
+ * @license GNU General Public License - http://www.gnu.org/copyleft/gpl.html
+ * @author Laurent Opprecht <laurent@opprecht.info>
+ * @translation Hans Haufe <hans@cappon.org>
+ */
+$strings['plugin_title'] = "Statisch";
+$strings['plugin_comment'] = "Statische html inhoud weergeven.;
```
AngelFQC, line 9: Missing trailing `"` in this string.

spacecabbie: Doh.... fixed.

PR #3285, opened 2020-06-07
change HOST to ADDR

spacecabbie:
ADDR is recommended no clue why and what the differance is exactly but for me HOST dus not work ADDR does.
also added ^ for https.


ywarnier:
The explanation here helps understand why: https://stackoverflow.com/questions/3812166/difference-between-remote-host-and-remote-addr
i.e. REMOTE_HOST can be a machine name, REMOTE_ADDR is an IP address.
Here we clearly want the IP address, so... PR accepted (better late than never)

PR #3329, opened 2020-06-23
3309

carlangas159:
Export de vcard non-fonctionnel 3309

Allow download vcard to a user with relationship.

### DIFF
--- a/main/social/vcard_export.php
+++ b/main/social/vcard_export.php
@@ -25,7 +25,10 @@ $userInfo = api_get_user_info($userId, true, false, true);
 
 /* Get the relationship between current user and vcard user */
 $currentUserId = api_get_user_id();
-$hasRelation = (int) (SocialManager::get_relation_between_contacts($currentUserId, $userId));
+$hasRelation =  SocialManager::get_relation_between_contacts(
+    $currentUserId,
+    $userId,
+    true);
 if ($hasRelation == 0) {
     /* if has not relation, check if are admin */
     api_protect_admin_script();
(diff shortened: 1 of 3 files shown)


main/social/vcard_export.php
```diff
@@ -24,6 +23,14 @@
 // Return User Info to vCard Export
 $userInfo = api_get_user_info($userId, true, false, true);
 
+/* Get the relationship between current user and vcard user */
+$currentUserId = api_get_user_id();
+$hasRelation = (int) (SocialManager::get_relation_between_contacts($currentUserId, $userId));
+if ($hasRelation == 0) {
```
jmontoyaa, line 29: Better to use:

```
if (0 === $hasRelation) {
}
```

main/social/vcard_export.php
```diff
@@ -24,6 +23,14 @@
 // Return User Info to vCard Export
 $userInfo = api_get_user_info($userId, true, false, true);
 
+/* Get the relationship between current user and vcard user */
+$currentUserId = api_get_user_id();
+$hasRelation = (int) (SocialManager::get_relation_between_contacts($currentUserId, $userId));
```
jmontoyaa, line 28: Don't add "(int)" better make sure that SocialManager::get_relation_between_contacts returns an int value doing the cast inside the function.

main/social/vcard_export.php
carlangas159: indeed, the function get_relation_between_contacts returns the id of the record or the USER_UNKNOWN constant, for security I set the int in the case that row ['id'] returns a string

main/social/vcard_export.php
carlangas159: indeed, the function get_relation_between_contacts returns the id of the record or the USER_UNKNOWN constant, for security I set the int in the case that row ['id'] returns a string

jmontoyaa: Yes, exactly.

PR #3344, opened 2020-06-30
#3079 Incorrect status tag in sections of a lesson.

carlangas159:
Hidding status from section. Will show only like tittle


AngelFQC:
No corrige el issue

PR #3347, opened 2020-07-01
HR managers did not receive a copy of the messages #3333

carlangas159:
Prevents error when "user_id" is not found but "id" if it exists.

Copy condition added to RH

### DIFF
--- a/main/inc/lib/message.lib.php
+++ b/main/inc/lib/message.lib.php
@@ -782,12 +782,9 @@ class MessageManager
                         get_lang('CopyOfMessageSentToXUser'),
                         $userInfo['complete_name']
                     ).' <br />'.$message;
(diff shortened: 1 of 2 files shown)


main/inc/lib/message.lib.php
```diff
@@ -782,6 +782,9 @@ public static function send_message_simple(
                         get_lang('CopyOfMessageSentToXUser'),
                         $userInfo['complete_name']
                     ).' <br />'.$message;
+                    if(!isset( $drhInfo['user_id']) and isset( $drhInfo['id'])){
+                         $drhInfo['user_id'] = $drhInfo['id'];
+                    }
```
jmontoyaa, line 787: That code is not needed. UserManager::getDrhListFromUser returns only "id"

Better to fix:

```
self::send_message_simple(
    $drhInfo['id'],
```

PR #3354, opened 2020-07-08
Check course limits to allow a new course creation

juan-cortizas-ponte:
[configuration.php](https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/install/configuration.dist.php) has two configurable course creation limit parameters:

**'[hosting_limit_courses](https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/install/configuration.dist.php#L91)'**: if set, sets the max number of courses
**‘[hosting_limit_active_courses](https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/install/configuration.dist.php#L102)’:** if set, set a maximum number of usable courses, defining a course as "hidden" will free room for new courses up to the hosting_limit_courses value

main/inc/lib/course.lib.php  [checkCreateCourseAccessUrlParam](https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/inc/lib/course.lib.php#L6955#L6969) function compares the total number of courses created on the platform with the value of the above parameters.

imagen

For **'hosting_limit_active_courses'** is incorrect because it not filter the hidden courses, always compares it value with the total number of courses.

Also, even if the course creation limits are broken, the course would have been created anyway because this function always return false.

imagen

imagen

This pull add the optional parameter $filterHiddenCourses = false
on checkCreateCourseAccessUrlParam, if not set or false check it will use function count_courses($access_url_id = null, $visibility = null) to get the numbers of courses created,if set to true it will use function countActiveCourses($urlId = null) to get the number of not hidden courses.

imagen

Also, checkCreateCourseAccessUrlParam returns a true value in case the course limit was broken

imagen

### DIFF
--- a/main/inc/lib/course.lib.php
+++ b/main/inc/lib/course.lib.php
@@ -69,8 +69,7 @@ class CourseManager
                 $_configuration,
                 $accessUrlId,
                 'hosting_limit_active_courses',
-                'PortalActiveCoursesLimitReached',
-                true
+                'PortalActiveCoursesLimitReached'
             );
             if ($return != false) {
                 return $return;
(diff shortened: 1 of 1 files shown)


main/inc/lib/course.lib.php
```diff
@@ -6952,16 +6953,22 @@ public static function getCoursesByUserCourseCategory($user_id)
      *
      * @return bool|string
      */
-    private static function checkCreateCourseAccessUrlParam($_configuration, $accessUrlId, $param, $msgLabel)
+    private static function checkCreateCourseAccessUrlParam($_configuration, $accessUrlId, $param, $msgLabel, $filterHiddenCourses = false)
```
jmontoyaa, line 6956: No es necesario agregar una nueva variable: $filterHiddenCourses 
$param ya tiene el tipo de configuración.

Mejor hacer algo así: 
```

private static function checkCreateCourseAccessUrlParam($_configuration, $accessUrlId, $param, $msgLabel)
    {
        if (isset($_configuration[$accessUrlId][$param]) && $_configuration[$accessUrlId][$param] > 0) {
            $num = null;
            switch ($param) {
                case 'hosting_limit_courses':
                    $num = self::count_courses($accessUrlId);
                    break;
                case 'hosting_limit_active_courses':
                    $num = self::countActiveCourses($accessUrlId);
                    break;
            }

            if ($num && $num >= $_configuration[$accessUrlId][$param]) {
                api_warn_hosting_contact($param);

                Display::addFlash(
                    Display::return_message($msgLabel)
                );

                return true;
            }
        }

        return false;
    }
```

juan-cortizas-ponte:
Modificado, gracias @jmontoyaa !

PR #3367, opened 2020-07-11
create_learningpath web service - refs BT#17453

sebastiendu:
Moved most learning path creation code from main/lp/learnpath.class.php to CourseBundle/Entity/CLp.php and CLpItem.php.


src/Chamilo/CoreBundle/Entity/Course.php
```diff
@@ -299,6 +406,119 @@ public function __toString()
...
+    {
+        return Database::getManager()->getRepository('ChamiloCoreBundle:Course');
+    }
+
+    /**
+     * @return Course|null
+     */
+    public static function getCurrentCourse()
+    {
+        return self::getRepository()->find(api_get_course_int_id());
```
jmontoyaa, line 422: Better not to mix legacy code (from api.lib.php), to this new entity classes...

NicoDucou:
Ce PR est inclus dans le PR https://github.com/chamilo/chamilo-lms/pull/3372 donc il sera revu directement dans cet autre PR qui contient d'autres Web Services en même temps.

jmontoyaa:
This PR has code already sent in #3372 and it has the same issues (database changes, core chamilo refactoring)

main/lp/lp_subscribe_users.php
```diff
@@ -159,7 +159,7 @@
 }
 $form->setDefaults($defaults);
 
-$currentUser = api_get_user_entity(api_get_user_id());
+$currentUser = api_get_user_entity();
```
AngelFQC, line 162: `api_get_user_entity` always needs an argument
https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/inc/lib/api.lib.php#L1798

src/Chamilo/CourseBundle/Entity/CLp.php
```diff
@@ -266,20 +276,240 @@ class CLp
...
+    /**
+     * @var Course
+     * @ORM\ManyToOne(
+     *     targetEntity="Chamilo\CoreBundle\Entity\Course",
+     *     inversedBy="learningPaths",
+     *     cascade={"persist", "remove"}
+     * )
+     * @ORM\JoinColumn(name="c_id", referencedColumnName="id")
+     */
+    protected $course;
```
AngelFQC, line 288: @jmontoyaa Can this association mapping for entities generate a change in the DB structure?

jmontoyaa: Yes, It will create FKs in the database.

main/lp/lp_subscribe_users.php
sebastiendu: This pull request also changes api_get_user_entity definition: the argument becomes optional - in cas it is not provided, the current logged-in user id is automatically substituted. As it is already the case with the other api_get_*_entity global functions : api_get_course_entity() returns the current course, api_get_session_entity() returns the current session, and so on.

src/Chamilo/CourseBundle/Entity/CLp.php
sebastiendu: Which should not create any problem, since only new Chamilo installations will get this new foreign key constraint (which is a security). Existing database will not be altered.

main/lp/lp_subscribe_users.php
sebastiendu: I should have splitted this too big PR in a sequence of smaller ones.

AngelFQC: > This pull request also changes api_get_user_entity definition: the argument becomes optional - in cas it is not provided, the current logged-in user id is automatically substituted. As it is already the case with the other api_get_*_entity global functions : api_get_course_entity() returns the current course, api_get_session_entity() returns the current session, and so on.

Ok. The diff is very long for api.lib.php and the file is not shown

src/Chamilo/CourseBundle/Entity/CLp.php
ywarnier: ...which leaves us with only the large changes to the core processes of Chamilo (even if only a change in form), which is a bit too risky to include in the code a few days before the release (we have to live with that with users support for months if there is something wrong).
The inclusion of this PR is thus postponed to after the 1.11.12 release.
I want to stress that this is not about bad code or anything (to the countrary). It is about the size of the changes and the corresponding risk.

ywarnier:
Implemented through a986f31 for 2.0

Issue #3191, opened 2020-04-18
Limit speed of user searches in social network


ywarnier:
In the social network and the message sending (/main/social/search.php?q=some+user+name and main/messages/new_message.php), the search for users is made in a way that each keystroke (javascript event onkeyup, I think), a search is sent in the users table to find the corresponding users.

This (the onkeyup) is too fast. It should at least wait one second before sending each request (wait for the accumulation of keystrokes for one second or more), otherwise the server gets flooded, risking the user to be blocked by firewalls.

On portals with a very high number of users (more than 10.000, for example), each search request can last several seconds to come back with all the matching results, which is slower than the rate of requests sent, which creates a queue which can overload a server.

Version 1.11.x of 20200417

AngelFQC:
In `message.ajax.php` exists a limit in the query when searching usernames... 
like `a=find_users&q=baggins&page_limit=10`

ywarnier:
OK, but the SQL search is still on all the users... So slowing down the reaction time is needed

ywarnier:
Or searching on less fields...

ywarnier:
Tested and approved

PR #3368, opened 2020-07-13
Limit speed of user searches in social network #3191

carlangas159:
Limit speed of user searches in social network #3191

Add delay to select2 option, set it to 1000 ms.

More info about select2 ajax options [https://select2.org/data-sources/ajax](https://select2.org/data-sources/ajax)

### DIFF
--- a/main/inc/lib/formvalidator/Element/SelectAjax.php
+++ b/main/inc/lib/formvalidator/Element/SelectAjax.php
@@ -75,6 +75,10 @@ class SelectAjax extends HTML_QuickForm_select
         $max = $this->getAttribute('maximumSelectionLength');
         $max = !empty($max) ? "maximumSelectionLength: $max, " : '';
 
+        // wait XX milliseconds before triggering the request
+        $delay = $this->getAttribute('delay');
+        $delay = !empty($delay) ? "delay: $delay, " : '';
+
         $html = <<<JS
             <script>
                 $(function(){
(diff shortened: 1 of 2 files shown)


main/inc/lib/formvalidator/Element/SelectAjax.php
```diff
@@ -75,6 +75,10 @@ public function toHtml()
         $max = $this->getAttribute('maximumSelectionLength');
         $max = !empty($max) ? "maximumSelectionLength: $max, " : '';
 
+        // wait XX milliseconds before triggering the request
+        $delay = $this->getAttribute('delay');
```
jmontoyaa, line 79: int casting missing

`$delay = (int) $this->getAttribute('delay');`

main/messages/new_message.php
```diff
@@ -149,6 +149,7 @@ function manageForm($default, $select_from_user_list = null, $sent_to = '', $tpl
                     [],
                     [
                         'multiple' => 'multiple',
+                        'delay' => '1000',
```
jmontoyaa, line 152: This should be probably an int instead of a text.

carlangas159: int casting missing

$delay = (int) $this->getAttribute('delay'); 

$delay will be 0 when return null, so change this validation

$delay = !empty($delay) ? "delay: $delay, " : '';

to

$delay = (0 !== $delay) ? "delay: $delay, " : '';

PR #3375, opened 2020-07-15
add 'Time spent/Time required' field on accordion lessons list

juan-cortizas-ponte:
Chamilo [has an option to activate an accordion view list](https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/install/configuration.dist.php#L440) for the lesson list.

If the option is enabled we will lose the "Time spent / Time required" field

**$_configuration['lp_category_accordion'] = false**

teacher view:

imagen

alumn view:

imagen

**$_configuration['lp_category_accordion'] = true**

teacher view:

imagen

alumn view:

imagen

This pull update the [/main/template/default/learnpath/list.tpl](https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/template/default/learnpath/list.tpl) template adding to the accordion view the "Time spent / Time required" fields (info_time_prerequisite)

**$_configuration['lp_category_accordion'] = true**

teacher view:

imagen

alumn view:

imagen

### DIFF
--- a/main/template/default/learnpath/list.tpl
+++ b/main/template/default/learnpath/list.tpl
@@ -221,10 +221,16 @@
                                             <th>{{ "PublicationDate"|get_lang }}</th>
                                             <th>{{ "ExpirationDate"|get_lang }}</th>
                                             <th>{{ "Progress"|get_lang }}</th>
+                                            {% if allow_min_time %}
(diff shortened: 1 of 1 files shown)


ywarnier:
Looks good.
The md-8 -> md-6 could be made dependent on that prerequisite time too, but I don't think that changes a lot in the display part, so all good from my point of view.

Issue #3299, opened 2020-06-15
Install: Hide admin password


ywarnier:
**Describe the solution you'd like**
During installation (step 5) the proposed password is show unhidden. Offer a possibility (eye icon) to hide it. By default, it should be visible as there is an automated password generated.

ywarnier:
Tested and approved. Thanks!

PR #3376, opened 2020-07-15
Install: Hide admin password #3299

carlangas159:
Install: Hide admin password #3299

Add fontawsome eye to show/hide password.


ywarnier:
The un-hidden field appears twice for now.

PR #3383, opened 2020-07-17
global and user Zoom meetings - refs BT#17288

sebastiendu:
Introducing 3 local tables to link remote Zoom API meetings to local
user, course and sessions and remote registrants to local users.
Better than using remote storage. Opens new possibilities.
Now works with "webhooks" to maintain local copies of Zoom objects.


ywarnier:
Looks good to me. Nice rewrite. Thanks!

PR #3393, opened 2020-07-21
Tests/Report by question broken for many types of tests #3177

carlangas159:
Tests/Report by question broken for many types of tests #3177 

Get all question of exercise and display in stats.php instead of showing the amount selected for the test


ywarnier:
Looks good. The user who reported will still have to confirm that this fixes the issue.

Issue #3394, opened 2020-07-22
Agenda: Repeated "All day" events show on full day span after first occurence


ywarnier:
**Describe**
When I created a repeated "All day" event in a course agenda, the first event is created great, but the others span the whole day (instead of appearing as an all day item at the top) in "Week" view.

**To Reproduce**
1. Create a course
2. Go to agenda
3. Click "Create event"
4. Give a name, leave the date/time alone, mark "All day" and mark "Repeat", leave it at "Daily" and mark the last repetition 2 or 3 days later
5. Save
6. Go to the "Week" view
7. Only the first event is marked as an "All day" event. The others are event that span from 00:00 to 23:59

**Expected behavior**
All the repetitions of an "All day" event should appear as "All day" events.

**Screenshots**
![agenda-form](https://user-images.githubusercontent.com/609839/88187350-ae223880-cc36-11ea-99a1-6ed1a37017f4.png)
![events](https://user-images.githubusercontent.com/609839/88187356-af536580-cc36-11ea-96b1-886e1d54d343.png)

Version 1.11.x of today.

carlangas159:
The way to evaluate the event for the whole day was not being taken correctly, therefore, after inserting the first event, the all_day field was left at 0
![image](https://user-images.githubusercontent.com/18097392/88218404-a0b08280-cc25-11ea-9a55-feffb7aee94d.png)

ywarnier:
Checked and approved. Thanks!

PR #3395, opened 2020-07-22
Agenda: Repeated "All day" events show on full day span after first occurence #3394

carlangas159:
Agenda: Repeated "All day" events show on full day span after first occurence #3394

From the second round, allDay changes from true to 1, as compared to the string true, it will not validate the date for its continuation.

Validation against 1 is added directly to compare against numeric value 1 that can be assumed to be true.

The database all_day field in table c_calendar_event is set to 1 correctly.

### DIFF
--- a/main/inc/lib/agenda.lib.php
+++ b/main/inc/lib/agenda.lib.php
@@ -246,7 +246,7 @@ class Agenda
     ) {
         $start = api_get_utc_datetime($start);
         $end = api_get_utc_datetime($end);
-        $allDay = isset($allDay) && ($allDay === 'true' or $allDay == 1) ? 1 : 0;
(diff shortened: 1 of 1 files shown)


main/inc/lib/agenda.lib.php
```diff
@@ -246,7 +246,7 @@ public function addEvent(
     ) {
         $start = api_get_utc_datetime($start);
         $end = api_get_utc_datetime($end);
-        $allDay = isset($allDay) && $allDay === 'true' ? 1 : 0;
+        $allDay = isset($allDay) && ($allDay === 'true' or $allDay == 1) ? 1 : 0;
```
AngelFQC, line 249: Don't use `or`. Use `&&`

PR #3400, opened 2020-07-23
Exercise result: Exporting pdf is impossible #3222

carlangas159:
Possible fix to
"Notice: ob_clean(): failed to delete buffer. No buffer to delete in /furanet/sites/11.chamilo.org/web/htdocs/main/exercise/exercise_show.php on line 942"

https://www.php.net/manual/en/function.ob-clean.php#Hcom114967

Check if had buffer.

### DIFF
--- a/main/exercise/exercise_show.php
+++ b/main/exercise/exercise_show.php
@@ -939,8 +939,7 @@ if ($show_results) {
 if ('export' === $action) {
     $content = ob_get_clean();
     // needed in order to mpdf to work
-    ob_clean();
-
+    if (ob_get_contents()) ob_clean();
     $params = [
         'filename' => api_replace_dangerous_char(
             $objExercise->name.' '.


carlangas159:
i can't reproduce this on my local chamilo (windows) and  into https://dev1.beeznest.com/.


I did several tests and I could not replicate the download of the wrong pdf. Although in a moment I noticed that it was not downloading correctly, 5 seconds later it opened the file correctly.

At the moment on 11.chamilo.org in the following link

https://11.chamilo.org/main/exercise/exercise_show.php?cidReq=NICODU&id_session=0&gidReq=0&gradebook=0&origin=&id=224&action=export&

"Notice: ob_clean (): failed to delete buffer. No buffer to delete in /furanet/sites/11.chamilo.org/web/htdocs/main/exercise/exercise_show.php on line 942"

There may be no buffer stored when trying to clean, so I check to see if the buffer exists.

It is still possible that the errors in 11.chamilo.org persist regarding:

"Warning: count (): Parameter must be an array or an object that implements Countable in /furanet/sites/11.chamilo.org/web/htdocs/vendor/mpdf/mpdf/classes/otl.php on line 5840"

content is a character string, at the moment it would be good to check if it is a problem by ob_clean

ywarnier:
As per https://www.php.net/ob_get_contents, this looks like a reasonnable attempt at fixing it...

Issue #3008, opened 2019-10-05
Incorrect time in Thematic Advance


rejuvenator:
**Describe**
The time displayed in Course Progress/Thematic Advance (and taken from Attendances) is wrong - it displays Greenwich Mean Time instead of my local time. Timezone on my server is set up correctly, and also the correct time is displayed elsewhere in the course, including Attendances and even the list of dates/times on the _Edit thematic advance_ page.

**To Reproduce**
1. Attendances: Create a new attendance list (repeated dates)
2. In Course Progress, add new thematic section, 
3. Create new thematic advance and select one of the dates.
4. Confirm and see different time in Course Progress/Thematic Advance (assuming you're not actually using GMT as your system time)

**Expected behavior**
Local (server) time displayed instead of GMT

**Desktop (please complete the following information):**
 - OS: Manjaro 18.1.0 KDE
 - Browser: Vivaldi / Opera / Chrome

**Server**
 - OS: [e.g. Linux Ubuntu 18.04/Windows 10]
 - Version of Chamilo 1.11.10
 - Version of PHP: 7.3

ywarnier:
Just to cut a bit the testing time... what timezone do you use?
It seems like the code of the thematic advance tool in general is local-time "aware", so we will need to dive deeper to understand the issue for you.
https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/course_progress/thematic_advance.php#L174

rejuvenator:
I use Central European time (Warsaw). The time shown in the thematic advance column is two hours earlier, which I think is GMT.

ywarnier:
Internal note: This would be in php.ini
```
date.timezone = 'Europe/Warsaw';
```

ywarnier:
Assigned to @NicoDucou for review

NicoDucou:
I set my date.timezone to Europe/Warsaw and it took me a bit of time to find where the problem was, because most of the dates are correct, it is just on the main Course progress tool homepage that the date is still incorrect (/main/course_progress/index.php?cidReq=GITHUB3008&id_session=0&gidReq=0&gradebook=0&origin=&action=thematic_details) The date on this page is different that all the other date.

![image](https://user-images.githubusercontent.com/6579268/89412472-74b5f680-d727-11ea-926c-8fe2df91d97e.png)

We can see in the picture above that the time is 11:47 but on the edition page it shows 13:47 as show in the picture below
[...]

PR #3401, opened 2020-07-23
Incorrect time in Thematic Advance - see #3008

carlangas159:
Incorrect time in Thematic Advance #3008

into course progress, index.php

    Prevent progress empty (show 0 %)

introductionSection.inc.php

    Prevent notice of undefined index of:

        $thematic_advance_info -> thematic_id /  start_date / content / duration
        $thematic_info -> title

    Double date setting increases the time at the scheduled course time.

### DIFF
--- a/main/course_progress/index.php
+++ b/main/course_progress/index.php
@@ -114,6 +114,11 @@ $(function() {
 </script>';
 
 $htmlHeadXtra[] = '<script>
+$(document).ready(function () {
+    if ($("#div_result").html() !== undefined && $("#div_result").html().length == 0) {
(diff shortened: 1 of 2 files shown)


main/course_progress/index.php
```diff
@@ -114,6 +114,11 @@
 </script>';
 
 $htmlHeadXtra[] = '<script>
+$(document).ready(function () {
```
AngelFQC, line 117: Mejor usar `$(function() {});` sin el `ready` porque ese método será removido en jQuery 3.0

Issue #3409, opened 2020-07-26
Learning Path View on Mobile Not working


aelalfey:
Dears @NicoDucou @ywarnier @AngelFQC , 

**Description:** When i view learning path on PC or Tablet it works just fine as in the picture below.

I have recreated the case at https://11.chamilo.org/main/lp/lp_controller.php?cidReq=MACCOURSE&id_session=0&gidReq=0&gradebook=0&origin=&action=view&lp_id=229&isStudentView=true

**PC View** everything works just fine.

![Bug 15](https://user-images.githubusercontent.com/33488577/88479124-411be500-cf4d-11ea-8775-a41dfacfe69b.jpg)

**Mobile View** right side is totaly invisible.

![Bug 16](https://user-images.githubusercontent.com/33488577/88479229-ecc53500-cf4d-11ea-8b49-25795712e496.jpg)

**One more thing** that i found while testing this on https://11.chamilo.org/main/lp/lp_controller.php?cidReq=MACCOURSE&id_session=0&gidReq=0&gradebook=0&origin=&action=view&lp_id=229&isStudentView=true When i tried to click on "Upload my assigment" button under "MAC assig. No 1" on the learning path, It is not working at all "the button is not even clickable"

![Bug 17](https://user-images.githubusercontent.com/33488577/88479320-8a206900-cf4e-11ea-893d-46ba517731fd.jpg)

Thank you once more for your help.
Regards

NicoDucou:
Hi @aelalfey 
> **One more thing** that i found while testing this on https://11.chamilo.org/main/lp/lp_controller.php?cidReq=MACCOURSE&id_session=0&gidReq=0&gradebook=0&origin=&action=view&lp_id=229&isStudentView=true When i tried to click on "Upload my assigment" button under "MAC assig. No 1" on the learning path, It is not working at all "the button is not even clickable"
> 
> ![Bug 17](https://user-images.githubusercontent.com/33488577/88479320-8a206900-cf4e-11ea-893d-46ba517731fd.jpg)

It works for me as you can see here :
![image](https://user-images.githubusercontent.com/6579268/88530678-0627b900-d002-11ea-9b09-9fd0488bc648.png)

We will now focus in this task on the other problem.

NicoDucou:
> **Description:** When i view learning path on PC or Tablet it works just fine as in the picture below.
> 
[...]

PR #3418, opened 2020-07-29
LP: Show item view when selecting an item in ToC #3409

AngelFQC:
Change the behaviour when clicking on an learning path item in a mobile view

merge


ywarnier:
Looks good to me. It makes more sense this way.

Issue #3437, opened 2020-08-10
GDPR fields titles in Terms and Conditons should be invisbile if there is no content


NicoDucou:
In the terms and conditions edition page (https://11.chamilo.org/main/admin/legal_add.php) there is a content Block at the top  and then there is a number of blocks corresponding to each point of the GDPR (Personal data collection, Personal data recording...). 

When connecting as a user that needs to validate the terms and conditions (because he has never validated it or because the terms and conditions have changed since last time), just after login the user is presented with the terms and conditions to be validated. You can see an exemple here https://11.chamilo.org/main/auth/inscription.php?legal connection with the account tcfr/tcfr (it's the terms and conditions in french that are presented to this user).

![image](https://user-images.githubusercontent.com/6579268/89765761-3ccbfc00-daf7-11ea-88f7-4d40767653e6.png)

Since there is no content in those blocks the title of the block should not be presented. Only show title and block if there is content.

NicoDucou:
It's all fixed for this task

PR #3439, opened 2020-08-10
GDPR fields titles in Terms and Conditons should be invisbile if there is no content

carlangas159:
GDPR fields titles in Terms and Conditons should be invisbile if there is no content - ref #3437


carlangas159:
> GDPR fields titles in Terms and Conditons should be invisbile if there is no content - ref #3437

image

ywarnier:
@NicoDucou this change seems rather simple. Could you check it on a system with GDPR items already set? (and also check why in Carlos' screenshot, there is an untranslated term - is this one missing?)

NicoDucou:
The change is correct and it works fine with already set GDPR content and without content.
For the translation, it's because of a missing get_lang so none of the title where translated.
I accept this PR and then I correct the get_lang issue.

NicoDucou:
Fixed with commit 107d15f9f8b739a3cca1183ebe746a79d43d207b

Issue #3428, opened 2020-08-05
Exercise result per question, incoherent data


NicoDucou:
In the Exercise report per question in the column "number of student that selected this" we only count student account but in the following columns we integrate all the result of all types of user.
So or we need to change the column "number of student that selected this" to show all type of user and not only student or in the following column just consider the students for the statistics.

Can be seen on https://11.chamilo.org/main/exercise/stats.php?cidReq=GITHUB3177&id_session=0&gidReq=0&gradebook=0&origin=&exerciseId=451
I have done the exercise 4 times as a teacher/admin so there is 0 in the number of student column but we can see statistics in the following columns : 

![image](https://user-images.githubusercontent.com/6579268/89437776-4ac0fc00-d748-11ea-98f3-1207d7b241ab.png)

tested on Chamilo 1.11.x of 05/08/2020

ywarnier:
We can change the title to "number of users having selected this one", and the filter should be to show all users always. This is not a critical report as in other cases (it helps design better questions, it doesn't define the future of a student), and it is important to include all users independantly from their roles (in future versions, a platform admin could be a simple student in a course).

NicoDucou:
It did not follow the comment of Yannick but now it takes only the students in all the calculation. So it is coherent now.

But doing this test I've seen another incoherence that I report here not to forget.
In the number of student the calculation take into account the number of student that selected one answer, but if the user did many attempt and all the time the same answer it only shows one time in the number of student that selected this answer but in the medium score it take into account all the attemps

Here is an explanation with a specific test case with 2 students (the first student did 2 attemps at the test and the second student did only 1 attemp).
![image](https://user-images.githubusercontent.com/6579268/89998309-a716a500-dc8d-11ea-8fa7-be23f0d4c95b.png)

carlangas159:
With the change that is made in the PR #3445 , the average is counted by all the attempts made by the students.

If the student performs the exercise 2 times, the progress bar shows 1 /? where the question is the number of students enrolled in the course and the average is calculated based on the 2 attempts.Con el cambio que se realiza, el promedio se cuenta por todos los intentos realizados por los estudiantes.

ywarnier:
[...]

PR #3445, opened 2020-08-12
Exercise: Exercise result per question, incoherent data. 

carlangas159:
Filters by students in class, although it takes all attempts made for the average - refs #3428

### DIFF
--- a/main/inc/lib/exercise.lib.php
+++ b/main/inc/lib/exercise.lib.php
@@ -3795,7 +3795,7 @@ EOT;
      * @param int    $exercise_id
      * @param string $course_code
      * @param int    $session_id
-     * @param bool   $onlyStudent   Filter only enrolled students
+     * @param bool   $onlyStudent Filter only enrolled students
      *
      * @return array
      */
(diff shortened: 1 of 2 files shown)


main/inc/lib/exercise.lib.php
```diff
@@ -3795,27 +3795,56 @@ public static function get_exercises_to_be_taken($course_code, $session_id)
...
+                )
+            ";
+        }
+
         $sql = "SELECT MAX(marks) as max, MIN(marks) as min, AVG(marks) as average
     		FROM $track_exercises e
-    		INNER JOIN $track_attempt a
+    		";
+        if ($onlyStudent == true) {
+            $sql .= $courseCondition;
```
ywarnier, line 3844: If you only use $courseCondition *if* $onlyStudent is true, then it shouldn't be necessary to build $courseCondition if $onlyStudent is not true (so you should build it *inside* the `if ($onlyStudent == true)` block.
I will do it myself to avoid delays, but I wanted to make sure you see it.

PR #3500, opened 2020-09-06
Master

00DarkGhost00:


ywarnier:
Not sure what this is, but it doesn't look like anything else than an attempt to see if we would accept any PR without checking...

PR #3519, opened 2020-09-14
Doodle : Set utc time at save's moment of end and start date

carlangas159:
In the course surveys, it is saved in local time instead of utc time. For this reason, the value of start_date and end_date is changed to adjust the utc time in Y-m-d H: i format since with seconds it saves null in the fields avail_from, avail_till of c_survey


ywarnier:
Looks good, thanks

PR #3556, opened 2020-10-11
Make it possible to replace a course document directly (rather than delete/create) - BT#17861

carlangas159:
Replaces the content of a file in course documents keeping its name. Simplifying the step of deleting the previous file, loading a new one and renaming it, in addition to maintaining relationships with the corresponding courses
image

### DIFF
--- a/main/document/document.php
+++ b/main/document/document.php
@@ -37,6 +37,9 @@ $parent_id = null;
 $lib_path = api_get_path(LIBRARY_PATH);
 $actionsRight = '';
 $action = isset($_REQUEST['action']) ? $_REQUEST['action'] : '';
+if(isset($_POST['currentFile'])){
+    $action = 'replace';
+}
 $allowUseTool = false;
 
 if ($allowDownloadDocumentsByApiKey) {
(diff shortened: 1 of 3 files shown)


main/inc/lib/document.lib.php
```diff
@@ -6614,6 +6660,154 @@ public static function getAllDocumentsCreatedByUser($userId)
...
+                                WHERE
+                                    c_id = $course_id AND
+                                    session_id = $sessionId AND
+                                    id = ".$documentId;
+                Database::query($sql);
+                $fileMoved = true;
+            }
+        }
+        // Checking inconsistency
+        if ($fileMoved ||
```
NicoDucou, line 6788: It should be && not || if I understand your logic well because we want the to action to be done to return true not only one option. For example if file_renamed_from_disk is true but fileMoved is false then it should not return true because something went wrong.

main/document/document.php
```diff
@@ -213,6 +216,82 @@
...
+                    $courseId,
+                    $group_properties,
+                    $documentInfo,
+                    true
+                );
+                // Check whether the document is in the database.
+                if (!empty($documentInfo)) {
+                    $file = $_FILES['file_'.$fileTarget];
+                    if ($documentInfo['filetype'] == 'file') {
+                        $deleteDocument = DocumentManager::writeContentIntoDocument(
```
NicoDucou, line 266: The variable should be named $updateDocument and not $deleteDocument it would be more logical. And then below use updateDocument instead of deleteDocument where ever this variable was used

carlangas159:
lang variable ReplaceFile

Issue #3596, opened 2020-11-01
Document : bug with cloud links (when trying to move it)


tanchon:
Hi,

In the configuration.php file, I enabled cloud links in document tool : $_configuration['enable_add_file_link'] = true;

In my document tool i created some folders : fold1 fold2...
... and i created some cloud  links : link1 link2 link3...

Everything is ok except when i want to move one of the link to another folder (from "document/" to "document/fold1/" for example.

I select link1 and try to move this file to "fold1/"  : this link is correctly moved but ... the link2 link is moved to !

When i make the opposite and I try to transfer the links1 from document/fold1 to document/,  it is the same issue : the link2 is moved to !

After these manipulations, another issue appear : the fold1 folder diasppears from my list of documents ! And when i try to create it, it is impossible : a warning message appears (the folder already exists).

**Expected behavior**
It should be possible to move the cloud links one by one, without any issue with folders.

**Server
Chamilo 1.11.12
PHP 7.3

This issue occurs with the "general" document tool and the "document" tool of the groups.
It occurs with every folder selected as a destination.
**Even when the domain names are different the issue can occur.**

Sometime, there is no problem : i couldn't determine any explanation.

Some screenshots
![Capture10](https://user-images.githubusercontent.com/24507698/97807208-d7a02680-1c5f-11eb-80c2-22f96e33ecfc.JPG)
![Capture11](https://user-images.githubusercontent.com/24507698/97807213-df5fcb00-1c5f-11eb-80a1-7ca5e3db3250.JPG)
![Capture12](https://user-images.githubusercontent.com/24507698/97807219-e686d900-1c5f-11eb-8189-4182710ba15a.JPG)

Thanks.

Tanchon.

carlangas159:
Hi @tanchon 

I have observed the problem, I have made a slight change that you can see in the pr https://github.com/chamilo/chamilo-lms/pull/3660, I comment that for the current links you must verify that the path field in c_document is correct, because the fault consists of

If the link is in folder 2 and you want to move it to folder 1, the path field would be folder-2 / folder-1 which is wrong, therefore in the path fields the folder-1 text must be eliminated so that reestablish.

![imagen](https://user-images.githubusercontent.com/18097392/100000433-9519c780-2d8f-11eb-8a5d-1915e18369ff.png)
I put an image to show you that you should delete in your database if it is the case.

tanchon:
Hi @carlangas159,

[...]

PR #3660, opened 2020-11-23
Document : Comment basename($document_to_move['path']) to prevent twice folde

carlangas159:
When $ _configuration ['enable_add_file_link'] = true; files can be uploaded to the cloud using the url. Then when you want to move to another folder, a failure occurs when placing the destination folder and also the current folder in the database.

In the DocumentManager :: updateDbInfo function it has been modified so that only the path to which it will be moved is set.
imagen
For previous events, the corresponding paths must be verified in the c_document table to set the "path" field in the correct location if there are problems with the current links
#3596

### DIFF
--- a/main/inc/lib/document.lib.php
+++ b/main/inc/lib/document.lib.php
@@ -6328,14 +6328,16 @@ class DocumentManager
      *
      * @author - Hugues Peeters <peeters@ipm.ucl.ac.be>
      *
-     * @param string $action   - action type require : 'delete' or 'update'
-     * @param string $old_path - old path info stored to change
-     * @param string $new_path - new path info to substitute
+     * @param string $action     - action type require : 'delete' or 'update'
+     * @param string $old_path   - old path info stored to change
+     * @param string $new_path   - new path info to substitute
+     * @param Int    $documentId - Id of specific document
      *
      * @desc Update the file or directory path in the document db document table
      */
-    public static function updateDbInfo($action, $old_path, $new_path = '')
+    public static function updateDbInfo($action, $old_path, $new_path = '', $documentId = 0)
     {
+        $documentId = (int) $documentId;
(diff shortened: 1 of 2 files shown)


main/inc/lib/document.lib.php
```diff
@@ -6328,14 +6328,16 @@ public static function deleteDocumentsFromSession($courseInfo, $sessionId)
      *
      * @author - Hugues Peeters <peeters@ipm.ucl.ac.be>
      *
-     * @param string $action   - action type require : 'delete' or 'update'
-     * @param string $old_path - old path info stored to change
-     * @param string $new_path - new path info to substitute
+     * @param string $action     - action type require : 'delete' or 'update'
+     * @param string $old_path   - old path info stored to change
+     * @param string $new_path   - new path info to substitute
+     * @param Int    $documentId - Id of specific document
```
jmontoyaa, line 6334: int sin mayúscula

main/inc/lib/document.lib.php
```diff
@@ -6363,6 +6365,9 @@ public static function updateDbInfo($action, $old_path, $new_path = '')
                           WHERE
                                 c_id = $course_id AND
                                 (path LIKE BINARY '".$old_path."' OR path LIKE BINARY '".$old_path."/%')";
+                if ($documentId != 0) {
+                    $query .= " AND id = $documentId";
```
jmontoyaa, line 6369: Mejor usar iid

jmontoyaa (changes requested):

main/inc/lib/document.lib.php
carlangas159: Muchas gracias, use id por que en principio, iid es parte de codigo viejo y no vi diferencias entre id e iid, para normalizarlo era mejor usar id. sin embargo, tomare en cuenta tu comentario.

Issue #3529, opened 2020-09-18
Accesses by user overview (Beta) on clicking page 2 going blank


kemoche:
**Describe**
Upon accessing the link "Accesses by user overview (Beta)" and after generating the report, on either clicking page 2 or change the number of row to see from 20 to 100/All, the page goes blank. The generated report vanishes.

**To Reproduce**
1. Login as Administrator ->Reporting ->Admin View ->Accesses by user overview (Beta)
2. Fill in the fields and Click on generate report
3. Change the number of rows to view ie from 20 to 500 or go to page 2
4. See error - blank page

**Expected behavior**
To see the full generated report

**Screenshots**
See attached screenshot.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser chrome, waterfox latest version

**Server
 - OS CentOS 7.8
 - Version of Chamilo: 1.11.12
 - Version of PHP: 7.2
![Access by user](https://user-images.githubusercontent.com/69464866/93632226-a93ef280-f9f5-11ea-99a2-25b8d2787cbd.jpg)

**Additional context**
It would be nice if I can also be able to download the data

carlangas159:
hi @kemoche 

I have reviewed the problem of why it is not displayed, I have already made an adjustment so that the table is displayed correctly, could you verify it? the related commit is at https://github.com/chamilo/chamilo-lms/pull/3661/commits/37ba6b62b3805d6063e9011ed528c11e19882419

However at the moment we could not attend to the extra functionality that you tell us

ywarnier:
The download should be registered as a feature request in another issue.

PR #3661, opened 2020-11-23
Adjustment for when you select a different amount of items per page when action is accessoverview

carlangas159:
In the report of accesses by user, when the amount of elements that will be displayed per page is changed, it does not appear.

An adjustment is made to remove $ form-> validate () since when you change the amount of elements it is a link and not a delivery method. However, it is evaluated that it contains dates for the date range, thus preventing an empty query from being made.

Because it is a query as a link because of the shape of the table, the elements necessary for the query are added in a form corresponding to SortableTable with the property set_additional_parameters.
To prevent inconsistencies in this way, it is evaluated if a variable exists by get method, if it does not exist but if it exists by post method then said variable will be taken. It also conforms to the get method for the getUserDataAccessTrackingOverview function

imagen

#3529

### DIFF
--- a/main/inc/lib/myspace.lib.php
+++ b/main/inc/lib/myspace.lib.php
@@ -3338,7 +3338,13 @@ class MySpace
      * @param int $sessionId
      * @param int $studentId
      */
-    public static function displayTrackingAccessOverView($courseId, $sessionId, $studentId)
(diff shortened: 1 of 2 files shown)


ywarnier:
I don't like much the assignation *to* `$_GET['something']`, but otherwise it looks good.

PR #3673, opened 2020-11-26
Documents: Use show_system_folders to show/hide default folders

### LINKED ISSUE #3621: Documents (in a course) : show system folders (and visibility of default multimedia folders)
Hi,

In the course settings, there is the "Show system folders" setting.

As it is configured, it is not very useful for the teacher. But i think it should be.
Indeed, it should be useful if this setting would have an effect on the "multimedia folders visibility". 

With Chamilo 1.11.12, only the admin platform can define it with "show default multimedia folders" setting in the admin panel. 

On my platform, i defined "show default multimedia folders" to "true" because it was a request from some teachers. 
But some others don't wan't to see these multimedia folders : if they set "Show system folders"  to "NO" it has no effect (the multimedia folders are visible).

**Describe the solution you'd like**
It should be great if there was a 3rd option for the admin platform with the  "show default multimedia folders" : yes, no, 'teacher choice'.
If the admin would choose the third choice, then the teachers could use "Show System folders" setting to define the visibility of multimedia folders.

**Additional context**
There is no consensus among teachers about  multimedia folders visibility. Teachers should have the choice.

Thanks.

Tanchon

AngelFQC:
Closes #3621

### DIFF
--- a/main/inc/lib/document.lib.php
+++ b/main/inc/lib/document.lib.php
@@ -4202,16 +4202,6 @@ class DocumentManager
      */
     public static function is_folder_to_avoid($path, $is_certificate_mode = false)
     {
-        $foldersToAvoid = [
-            '/HotPotatoes_files',
-            '/certificates',
-        ];
-        $systemFolder = api_get_course_setting('show_system_folders');
-
-        if ($systemFolder == 1) {
-            $foldersToAvoid = [];
-        }
-
         if (basename($path) == 'css') {
             return true;
         }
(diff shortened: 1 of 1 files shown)


tanchon:
Thanks @AngelFQC  !

It is more logical now 
1- if the admin platform decide to hide system folders, teachers haven't the choice
2- if admin platform decide to show system folders, teachers have the choice.

May be that it could be relevant to hide the setting in the infocours.php in the case 1 (because in this case, whatever the choice of the teacher is, it will change nothing). But it's a detail and the adjustment is perfect from my point of view !

Thanks.

Tanchon.

ywarnier:
Merged, including additional suggestion by @tanchon 
Thanks!

PR #3676, opened 2020-11-26
3656 -  Bug in site settings

lcubas:


ywarnier:
This looks OK. Good job.
However, because it modifies a method used in 52 different methods which in turn are called from many more locations, I feel like it is too late to safely include in 1.11.14.

I will leave this issue opened for now and merge this PR post-1.11.14.
In researching the impact of this, I have looked for 
 - the original inclusion of this lib in Chamilo (14 years ago) https://github.com/chamilo/chamilo-lms/blame/2a2f5d5af5731e9dcaba392d2ace77aedef2063d/main/inc/lib/pear/HTML/Common.php#L140 (pretty similar to its common form)
 - all the changes since then (only a change from ENT_QUOTES to ENT_COMPAT)
 - the changes to the original method: http://svn.php.net/viewvc/pear/packages/HTML_Common/trunk/Common.php?revision=278215&view=markup line 144 (no logic change detected)
 - the new version of this library: https://github.com/pear/HTML_Common2/blob/trunk/HTML/Common2.php#L223 (no logic change from the original)

Nothing suggests that this method has been a problem to others, which makes me doubt the necessity of applying this patch here.

ywarnier:
This change is too generic. It affects other forms of Chamilo where the filtering is different.

PR #3677, opened 2020-11-26
Quiz :  Lock back in browser when quiz_prevent_backwards_move is active

carlangas159:
When the option quiz_prevent_backwards_move is active, it prevents you from returning to the previous question when you are solving an exercise, which is correct.

However it can create confusion for students when they finish the exam and press back on the browser, they create another attempt which can create confusion for the student.

It seeks to block the back action of the browsers to prevent this from happening from the end of the exercise page and the option quiz_prevent_backwards_move is activated

NOTE: Validation is required since normally the back action of the browser should not be blocked (@AngelFQC )

### DIFF
--- a/main/exercise/exercise_result.php
+++ b/main/exercise/exercise_result.php
@@ -24,10 +24,13 @@ $this_section = SECTION_COURSES;
 
 api_protect_course_script(true);
 $origin = api_get_origin();
-
+$clearExercise = false;
 /** @var Exercise $objExercise */
 if (empty($objExercise)) {
     $objExercise = Session::read('objExercise');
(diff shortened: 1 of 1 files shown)


ywarnier:
I am intentionally leaving this PR open as this method goes against our normal logic (we avoid affecting normal browser behaviour). The recommendation (also given in the issue) is to only use this option with "single attempt" exercises.

ywarnier:
I tried it in many ways but couldn't get this feature to work as documented. I always go back one page and get the normal ("You already answered the question") message. Abandonning this PR as carlangas159 no longer works with us on this project.

PR #3687, opened 2020-12-01
check if the stored progress is higher than the new value

juan-cortizas-ponte:
Change on main/lp/learnpath.class.php to avoid updating the progress value with a lower value than the current one.

Changes on function save_last($score x null) to get the current stored progress value and compare it with the new value, if the new value is lower than the current one it does not save the progress

imagen

### DIFF
--- a/main/lp/learnpath.class.php
+++ b/main/lp/learnpath.class.php
@@ -4964,6 +4964,22 @@ class learnpath
                 $progress = $score;
             }*/
 
+            $storedProgress = self::getProgress(
+                $this->get_id(),
+                $userId,
+                $course_id,
+                $this->get_lp_session_id()
+            );
+
+            // Check if the stored progress is higher than the new value
+            if ($storedProgress >= $progress) {
+                if ($debug) {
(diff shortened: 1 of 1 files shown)


jmontoyaa:
This should be change only if "score as progress" is enabled

juan-cortizas-ponte:
updated

imagen

ywarnier:
@juan-cortizas-ponte, what @jmontoyaa means is that the calculation of the current progress should not be done (because it can be a heavy operation) unless we are in the case of score-as-progress.
I'm fixing it before merging. Don't worry.

juan-cortizas-ponte:
@ywarnier ok, thanks!

ywarnier:
Not sure why Github doesn't detect this as merged, but it has been squashed into one single commit and merged.

ywarnier:
Here: https://github.com/chamilo/chamilo-lms/commit/38e31f6712ff763ea5bc07c639e7b4829db08782
I also added one more commit with a little more info in the error log.

PR #3688, opened 2020-12-01
Updated LDAP config file location/name

lonesomewalker:
Corrected wrong filename where to edit LDAP config.

There is no main/auth/external_login/ldap.conf.php
Instead we have auth.conf.php

Also corrected some typo


ywarnier:
Thanks!

PR #3711, opened 2020-12-14
Fix to show theme for each course

fabiowoj:
The select theme for specific course not loading. The platform theme still appear. This fix, solve that bug.


ywarnier:
Thanks for reporting and offering a patch, @fabiowoj (and sorry for the delay reviewing it)
I have not taken your patch because it disables most of the api_get_visual_theme() optimisation. Rather, I have made a special case for when inside a course.
Submitted as c165bc5b58

Issue #2861, opened 2019-03-22
Survey displays question numbers


xam1234:
When you create survey all questions have numbers, this should be optional or turned off. Old versions didn't have this.

![f1](https://user-images.githubusercontent.com/32258255/54824732-10b69d80-4cac-11e9-88ef-772862aab509.jpg)

Server: 11.chamilo.org (22.03.2019)

ywarnier:
This requires a new field `display_question_number INT default 1` in the c_survey table.

ywarnier:
Works well for me

PR #3713, opened 2020-12-14
2861 - Survey displays question numbers

lcubas:


AngelFQC (changes requested):
Missing migration to add display_question_number to CSurvey

ywarnier:
#2861

PR #3738, opened 2021-01-05
Remove code that checks for Flash client

bigbluemeeting:
As the Flash client is no longer used and Flash has been deprecated in general there is no need for this code which checks for the client.


bigbluemeeting:
Closing this in preference for a more comprehensive request that removes all code related to multiple interfaces.

https://github.com/chamilo/chamilo-lms/pull/3739

PR #3739, opened 2021-01-05
Remove all code referencing Flash interface

bigbluemeeting:
Remove all code referencing Flash/multiple interface options as Flash has reached EOL and the client is to be removed from BigBlueButton.

- Remove widgets to configure choice from the interface.
- Remove checks for multiple interfaces.
- Remove options to allow various user categories to choose the interface.
- Remove language keys that are relevant to UI elements removed.


jmontoyaa:
Thanks @bigbluemeeting !

PR #3860, opened 2021-04-30
Gradebook display certificate - Prevent that the generate and delete certificate buttons are overwriten 

juan-cortizas-ponte:
Concatenates variable $actions to prevent that the generate and delete certificate buttons are overwriten if the custom certificates are allowed

imagen


jmontoyaa:
:+1: Nice!

PR #3871, opened 2021-05-11
WIP - Quiz: Remove usage of id in favor of iid in all quiz tables - refs BT#18452

ywarnier:
…#18452


ywarnier:
> Removing the $id property in entities generates a DB change

Good point. I have restored them. I don't think they will be affecting anything now, as having them removed still let the system work.

PR #3876, opened 2021-05-19
Fix XSS Breach in forum_posts

renaudlemaire:
A customer of us found a XSS Breach on forum posts using javascript


renaudlemaire:
updated

PR #3877, opened 2021-05-20
Cblue forum posts xss

renaudlemaire:
Fixing XSS Breach in forum

### DIFF
--- a/main/lp/lp_subscribe_users_to_category.php
+++ b/main/lp/lp_subscribe_users_to_category.php
@@ -220,6 +220,7 @@ if ($allowUserGroups) {
             $em->persist($category);
             $em->flush();
         }
+
         header("Location: $url");
         exit;
     }
@@ -312,6 +313,12 @@ if ($formUsers->validate()) {
         Display::addFlash(Display::return_message(get_lang('Updated')));
     }
 
+    header("Location: $url");
+    exit;
+}
+
+if ($form->validate()) {
+    $values = $form->getSubmitValues();
     // Subscribing groups
     $groups = isset($values['groups']) ? $values['groups'] : [];
     $groupForm = isset($values['group_form']) ? $values['group_form'] : [];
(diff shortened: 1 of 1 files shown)


main/forum/forumfunction.inc.php
```diff
@@ -3835,8 +3835,8 @@ function store_reply($current_forum, $values, $courseId = 0, $userId = 0)
             $table_posts,
             [
                 'c_id' => $courseId,
-                'post_title' => $values['post_title'],
-                'post_text' => isset($values['post_text']) ? ($values['post_text']) : null,
+                'post_title' => Security::remove_XSS($values['post_title']),
```
jmontoyaa, line 3838: There's no need to add Security::remove_XSS before the insert to the DB

main/forum/forumfunction.inc.php
```diff
@@ -4160,8 +4160,8 @@ function store_edit_post($forumInfo, $values)
 
     // Update the post_title and the post_text.
     $params = [
-        'post_title' => $values['post_title'],
-        'post_text' => $values['post_text'],
+        'post_title' => Security::remove_XSS($values['post_title']),
```
jmontoyaa, line 4163: The same here, Security::remove_XSS should be added before displaying the content.

main/forum/forumfunction.inc.php
renaudlemaire: Why keeping XSS'ed html in database?
If is not displayed, it's not useful, so not needed in db.

jmontoyaa: Security::remove_XSS uses HtmlPurifier and it's very agressive. In some cases it will remove valid HTML content.

It's better to use remove_XSS before displaying because we can edit the "aggressiveness" of the clean in order to allow some content.

renaudlemaire: Seems Legit. Thks.
I cleaned the branch and I made a new pull request #3884

ywarnier: @renaudlemaire We do this in general in Chamilo, because something stored in the database is not necessarily shown on a web page (it might get processed in a different way, like exporting in PDF, CSV, XML, etc). HTMLPurifier does clean it but sometimes we just might want the original post (even if only in terms of being able to audit it afterwards).
All output from the database to a web page should be filtered against XSS however... always (if you find a place where it's not, it's a mistake and probably a vulnerability).

renaudlemaire: @ywarnier Understood. Thanks.

PR #3878, opened 2021-05-20
force remove XSS on content

renaudlemaire:
No real XSS problem but malicious code is saved in database


renaudlemaire:
Viewed in another pull request. We want to keep the full html code received in database.

Issue #3839, opened 2021-04-04
Activating captcha renders the lost password function useless


lonesomewalker:
**Describe**
When activating the use of captcha, it is not possible to request a password because the captcha image is not shown.

**To Reproduce**
1. activate captcha in the admin backend
2. log out
3. click on "lost password"
4. see no captcha

Chamilo latest

ywarnier:
Checked and approved.

PR #3888, opened 2021-05-26
Fix captcha image is not displayed in lost password and inscription pages - refs #3839

christianbeeznest:


ywarnier:
This has been merged through a squash in 8d881fee0a (which apparently doesn't keep the original author, sorry about that).

Issue #3746, opened 2021-01-11
Greek Flag is not the right one


makatsis:
**Describe**
As you can see at the following screenshot the greek flag is not the right one. Chamilo 1.11.14 

<img width="643" alt="Στιγμιότυπο 2021-01-11, 2 30 53 μμ" src="https://user-images.githubusercontent.com/2492623/104183087-1863ab00-541a-11eb-9746-6f43314922eb.png">

ywarnier:
![greek-flag](https://user-images.githubusercontent.com/609839/168803222-4cf811df-d2ba-401b-a7d7-7275b358603c.png)

Greek flag is now correct

PR #3891, opened 2021-05-29
Fix greek flag is not the right one - refs #3746

christianbeeznest:


AngelFQC:
Aplicado por cherry-pick de 6937f358fd47ebbe644971d1ffb993a335ee67e3

PR #3900, opened 2021-06-04
Fix possible injection using HTTP_X_FORWARDED_FOR

renaudlemaire:
Not really dangerous but it is possible to inject content in HTTP_X_FORWARDED_FOR and it displays.
* check existance of REMOTE_ADDR in $_SERVER 
* display ip only if valid ip


jmontoyaa:
@renaudlemaire thanks!

PR #3904, opened 2021-06-09
Fix ajax user manager

binote00:
Fix case in get_user_popup when courseid is empty and user is anonymous


jmontoyaa:
It looks that this PR is for Chamilo 1.11.x.

Please send a PR to the 1.11.x branch instead of master.

PR #3905, opened 2021-06-09
fix ajax user manager

binote00:
Fix case in get_user_popup when courseid is empty and user is anonymous.

Avoid to return user information when not logged in. Before that it was possible to get user information with such a curl :

curl -H 'X-Requested-With: XMLHttpRequest' "https://platform_url/main/inc/ajax/user_manager.ajax.php?a=get_user_popup&user_id=23"


ywarnier:
Looks good to me. Sorry for the wait and thanks for the contribution!

PR #3909, opened 2021-06-10
Couse log - Option to set default extra fields as columns

juan-cortizas-ponte:
With this PR we can set some user extra fields to show automatically as columns on main/tracking/courseLog.php.

imagen

To achive it we need a new configuration.php parameter:

`$_configuration['course_log_default_extra_fields'] = [1 => []];`

For example, to show as a default column the user extra field with id 15 and 37 on site 1

`$_configuration['course_log_default_extra_fields'] = [1 => [15, 37]];`

Same config plus site 3 and extra fields with id 53 and 55:

`$_configuration['course_log_default_extra_fields'] = [1 => [15, 37], 3 => [53, 55]];`

Changes on main/inc/lib/tracking.lib.php:

1. Changed display_additional_profile_fields() to display_additional_profile_fields($exclude = []) to pass some extra fields to filter
2. Get on get_user_data($from, $number_of_items, $column, $direction, $conditions = []) the value for the the extra fields configuered as columns  at $_configuration['course_log_default_extra_fields']

Changes on main/install/configuration.dist.php:

1. Manage the $_configuration['course_log_default_extra_fields'] to show them as columns

### DIFF
--- a/main/install/configuration.dist.php
+++ b/main/install/configuration.dist.php
@@ -221,6 +221,12 @@ $_configuration['system_stable'] = NEW_VERSION_STABLE;
 //);
 // Course log - Default columns to hide
 //$_configuration['course_log_hide_columns'] = ['columns' => [1, 9]];
+// Course log - User extra fields to show as columns for default
+// For example, to show as a default column the user extra field with id 15 and 37 on site 1:
+// [1 => [15, 37]];
+// Same config plus site 3 and extra fields with id 53 and 55:
+// [1 => [15, 37], 2 => [53, 55]];
+//$_configuration['course_log_default_extra_fields'] = [1 => []];
 // Unoconv binary file
 //$_configuration['unoconv.binaries'] = '/usr/bin/unoconv';
 // Proxy settings for access external services
--- a/main/install/configuration.dist.php
+++ b/main/install/configuration.dist.php
@@ -222,11 +222,7 @@ $_configuration['system_stable'] = NEW_VERSION_STABLE;
 // Course log - Default columns to hide
 //$_configuration['course_log_hide_columns'] = ['columns' => [1, 9]];
 // Course log - User extra fields to show as columns for default
-// For example, to show as a default column the user extra field with id 15 and 37 on site 1:
-// [1 => [15, 37]];
-// Same config plus site 3 and extra fields with id 53 and 55:
-// [1 => [15, 37], 3 => [53, 55]];
-//$_configuration['course_log_default_extra_fields'] = [1 => []];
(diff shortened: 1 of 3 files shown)


main/tracking/courseLog.php
```diff
@@ -206,6 +206,32 @@ function(index) {
 Session::write('additional_user_profile_info', $userProfileInfo);
 Session::write('extra_field_info', $extra_info);
 
+$defaultExtraFields = [];
+$defaultExtraFieldsFromSettings = [];
+$defaultExtraFieldsFromSettings = api_get_configuration_value('course_log_default_extra_fields');
+if (isset($defaultExtraFieldsFromSettings)) {
+    $default_extra_fields = $defaultExtraFieldsFromSettings;
```
jmontoyaa, line 213: Use camelCase for  $default_extra_fields and $user_array

main/install/configuration.dist.php
```diff
@@ -221,6 +221,12 @@
 //);
 // Course log - Default columns to hide
 //$_configuration['course_log_hide_columns'] = ['columns' => [1, 9]];
+// Course log - User extra fields to show as columns for default
+// For example, to show as a default column the user extra field with id 15 and 37 on site 1:
+// [1 => [15, 37]];
```
jmontoyaa, line 226: Site 1 ? Site 3? What's that?

The configuration should be more clear for example:

```
$_configuration['course_log_default_extra_fields'] = [
    'extra_fields' => [
        'variable_test', 
        'variable_demo'
    ]
];
```
If the configuration is an array it should start with key of type string, in the example above the key is 'extra_fields'.

Or in this example : 

`$_configuration['course_log_hide_columns'] = ['columns' => [1, 9]];`

The key is "columns" otherwise there are conflicts with the Access url.

jmontoyaa (changes requested):

main/tracking/courseLog.php
juan-cortizas-ponte: Variables names changed to camelCase

main/install/configuration.dist.php
juan-cortizas-ponte: Change the configuration to :

`$_configuration['course_log_default_extra_fields'] = ['extra_fields' => [66, 67]];`

jmontoyaa: Use extra field variables instead of extra field ids, it will be more easy to edit later.

juan-cortizas-ponte: Fixed!: https://github.com/chamilo/chamilo-lms/pull/3909/commits/242c5e90b457b93d99fd7c65eb70d5f555de90ae

PR #3943, opened 2021-08-12
Fixes to LRS

AngelFQC:


ywarnier:
Please add the db structure and changes directly to the 2.0 install & migrate codebase, so the plugin is only an activation and can be better integrated in 2.1.
Ideally, this would move from plugin/xapi/ to another folder, but that's not critical for 2.0.

ywarnier:
This PR is still pending for merging into the master branch. Only a few days left before 2.0.

AngelFQC:
Closing this one: every change it carries is already in `master`, and the part that was still missing has been fixed separately.

The plugin moved to `public/plugin/XApi/` and its LRS tables became core entities (`src/CoreBundle/Entity/XApi*.php`, created by `Version20240515094800`), which is what @ywarnier asked for in January 2024. Change by change:

| Change in this PR | Status in `master` |
|---|---|
| `routing.xml`: `postStatement` → `postStatements` | applied (`7902b32cf57`) |
| `StatementsFilterFactory`: `$filter->cursor(...)` | applied (`37aede9d725`) |
| `StatementRepository`: verb filter + limit/cursor | applied (`37aede9d725`) |
| `StatementGetController`: formatting | applied |
| `Verb.orm.xml`: `display` nullable | equivalent in `XApiVerb::$display` (`nullable: true`) |
| `StatementObject.orm.xml`: cascade on `members` | equivalent in `XApiObject::$members` (`cascade: ['persist', 'remove']`) |
| `README.md`: `xapi_verb` DDL | obsolete — the tables are created by a migration now |
| `composer.json`: `php-xapi/*` bumps | not portable, see below |

The `composer.json` part cannot be brought over. Those packages are not merely unmaintained, they are incompatible by constraint with the 2.0 stack (PHP 8.2/8.3, Symfony 6.4/7, Doctrine ORM 3):

- `php-xapi/symfony-serializer` — last release v2.1.0 (2019-03-16), `dev-master` last touched 2021-03-26, requires `php ^7.1` and `symfony/serializer ^4.4 || ^5.2`
- `php-xapi/client` — v0.6.0 (2019), requires `php ^5.6 || ^7.0` and the abandoned `php-http/message-factory`
- `php-xapi/repository-doctrine` — v0.3.0 (2016), requires `doctrine/common ~2.4`
- `php-xapi/repository-api` — v0.3.1 (2020), requires `ramsey/uuid ^2.8 || ^3.0`
- `php-xapi/lrs-bundle` — not published on Packagist at all

That is why the dependencies were dropped when the plugin was ported, and the vendored copy under `public/plugin/XApi/php-xapi/` is dead code today: it references `Xabbuh\XApi\*` classes that are not installed.

What this PR was actually fixing did survive, though, in a different form. `StatementsController::get()`/`post()`/`put()` had been rewritten to keep statements in `$_SESSION`, so:

- statements did not survive the PHP session,
- they were only visible to the user who wrote them,
- and `XApiPlugin::fetchStatementsByRegistration()` — which the TinCan report calls over HTTP, server-to-server, without the learner's session cookie — always got back an empty list, so the teacher's report showed "No results found" whenever `lrs_url` pointed at the internal LRS.

That is now implemented natively on the core `XApi*` entities, with no php-xapi dependency: statements are persisted and queryable, `GET /statements` supports the 1.0.3 filters (`agent`, `verb`, `activity`, `registration`, `since`, `until`, `ascending`, `limit` + a `cursor`/`more` paging pair), and voiding follows the spec. Since those columns are now actually read on every request, a `V300` migration indexes them (`registration`, `verb.id`, `activity_id`, `referenced_statement_id`, `mbox`, `account_name`, `stored`).

Two mapping fixes were needed along the way, and they are the same class of bug this PR reported:

- `XApiContext` needed `cascade: ['persist', 'remove']` on its four `contextActivities` collections — without it, any statement carrying `contextActivities` failed to persist. This is the counterpart of the `<cascade-all />` this PR added to `StatementObject.orm.xml#members`.
- `XApiStatement` had no `setId()`, even though its identifier is `GeneratedValue(strategy: 'NONE')` and therefore has to be assigned by the caller.

Thanks for the patience on this one.

PR #3947, opened 2021-08-17
Editor : Improve upload when image file has the same name - BT18997

christianbeeznest:


NicoDucou:
All good for me at a functional level. Only needs a code revision/validation to be able to include it in Chamilo.

main/inc/ajax/document.ajax.php
```diff
@@ -176,7 +176,7 @@
                 '',
                 '',
                 0,
-                '',
+                'rename',
```
ywarnier, line 179: Aquí estás cambiando el comportamiento por defecto de Chamilo (warn vs rename) y no entiendo bien porqué.

ywarnier: Mmmm OK, I see in BT#18997 that this was a requirement.

PR #3976, opened 2021-09-16
BBB plugin - Option to close all meetings on cron cron_close_meeting.php process

juan-cortizas-ponte:
Add a new configuration.dist.php option to allow to close al meetings on a multi url campus

Changes on main/install/configuration.php -> **multiple_urls_cron_apply_to_all** new option parameter:
imagen

Changes on plugin/bbb/lib/bbb.lib.php -> function **getActiveSessions($allSites = false)**: 

imagen

Changes on plugin/bbb/cron_close_meeting.php -> use **multiple_urls_cron_apply_to_all** configuration value:

imagen

### DIFF
--- a/plugin/bbb/CHANGELOG.md
+++ b/plugin/bbb/CHANGELOG.md
@@ -1,3 +1,7 @@
+Version 3.0 - 2021-10
+---------------------
+* Add option to close all rooms on all campus on a multi-url enviroment
+
 Version 2.9 - 2021-08
 ---------------------
 * Remove interface option (HTML5/Flash)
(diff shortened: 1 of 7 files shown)


ywarnier:
The new configuration is a bit generic: it suggests it applies to all CRON processes, while it only applies to plugin/bbb/cron_close_meeting.php. So someone who doesn't use BBB could change this setting, thinking that it will apply to all CRON processes, and it would do nothing.
I would rename it directly to 'plugin_bbb_multiple_urls_cron_apply_to_all', but then that would be related to the plugin only and so it should be in the plugin itself.

Could you (please) change this PR to:
- add this setting as a new plugin setting for BBB
- add the information about the change to be done in the database in the changelog of the plugin
- query the plugin setting rather than api_get_configuration_value()

Sorry for the additional work, but I really think it would be confusing otherwise.

plugin/bbb/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+Version 3.0 - 2021-10
```
ywarnier, line 1: Should be 2.10. Not a major new version here.

ywarnier (changes requested):
Change version to 2.10 and check the observation on multi-urls

PR #4003, opened 2021-09-29
Tool intro: Add course introduction (partial) - refs BT#19044

christianbeeznest:


jmontoyaa:
Close as demanded by @christianbeeznest

PR #4020, opened 2021-10-14
User: Improve terms and conditions, fix warning errors and auto-register - refs BT#19044

christianbeeznest:


src/CoreBundle/Repository/LegalRepository.php
```diff
@@ -58,4 +61,97 @@ public function findOneByTypeAndLanguage(int $typeId, int $languageId)
...
+                $qb->expr()->eq('l.language_id', $languageId)
+            )
+            ->setMaxResults(1)
+            ->orderBy('l.version', Criteria::DESC)
+            ->getQuery()
+            ->getOneOrNullResult()
+            ;
+        if (!empty($result)) {
+            $version = explode(':', $result);
+            return $version[0];
```
ywarnier, line 152: Should be `return (int) $version[0];` to make sure it's an integer

PR #4032, opened 2021-10-19
Add helper to get legacy configuration values

AngelFQC:


jmontoyaa:
The code looks good, but I think is not a good idea to mix legacy and v2 settings. Better to move that config in the TicketSettingsSchema

PR #4033, opened 2021-10-19
Lp: Fix update progress inside lp view - refs BT#19283

christianbeeznest:


NicoDucou:
It seems all good to me and works fine in my test cases, but I leave it to Yannick to revise the PR before merging it.

PR #4041, opened 2021-10-25
Session cookie: Option to set samesite:None parameter

juan-cortizas-ponte:
Users using browsers with the Chromium engine (Chrome, Edge, ...) with chamilo campuses embedded in third-party sites via iframes (e.g., proctorized exam platforms) fail to login in.

This is due to changes in behavior for cross-site cookies introduced by Google in 2020 that require session cookies to be specified with the samesite parameter to None and secure to true.

https://www.chromium.org/updates/same-site
https://developers.google.com/search/blog/2020/01/get-ready-for-new-samesitenone-secure

To check this behavior create an html file with an iframe indicating in the src a Chamilo campus and open it with a browser with Chromium engine.

This pull add a new configuration parameter called "allow_session_samesite_none_cookie_parameter" (I really don´t like this name, any suggestions?), if its true adds to the session cookie the samesite:None and secure:True parameters

### DIFF
--- a/main/inc/lib/chamilo_session.class.php
+++ b/main/inc/lib/chamilo_session.class.php
@@ -87,7 +87,7 @@ class ChamiloSession extends System\Session
         //session ID in the cookie is only readable by the server
         ini_set('session.cookie_httponly', 1);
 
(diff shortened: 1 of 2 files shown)


ywarnier:
Ping @juan-cortizas-ponte

PR #4050, opened 2021-10-28
Reporting: Add lp stats export to pdf - refs BT#19310

christianbeeznest:


ywarnier:
This looks OK. Some queries could be moved to "less legacy" syntax, but for a PDF export, I think we're good. We can improve on this later on.

PR #4063, opened 2021-11-09
detection of user password encryption from database, grant login desp…

renaudlemaire:
Keeping password encrypted in md5 was an issue for a customer of us.
Here are small changes to grant login using password encryption from database and re-encrypt password using configured encryption

### DIFF
--- a/main/inc/lib/usermanager.lib.php
+++ b/main/inc/lib/usermanager.lib.php
@@ -169,7 +169,7 @@ class UserManager
         $result = false;
 
         $detectedEncryption = self::detectPasswordEncryption($encoded, $salt);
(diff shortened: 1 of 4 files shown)


ywarnier:
Nice @renaudlemaire! Given the overhead at login time, I think it would be even nicer to put that under a configuration option (you can see how to define and use one here, for example: https://github.com/chamilo/chamilo-lms/pull/4041/files). Something like `$_configuration['password_conversion'] = false;` (we always set them to false in the definition, even though they're commented out).
Would you be up to add that additional change?

renaudlemaire:
here it is

PR #4064, opened 2021-11-10
Reporting: Add option to hide skills before exporting to pdf - refs #19335

christianbeeznest:


christianbeeznest:
Variabla a traducir  HideSkills = Hide Skills

PR #4086, opened 2021-12-01
Prevent admin@example.com on notifications

LudiscapeXApi:
If email is admin@example.com use the admin email


ywarnier:
I would need more details here... When is the admin e-mail supposed to be admin@example.com ?

ywarnier:
Reshaped a little and sent in 687673522a614d310e99e73e99164223432e7b41

PR #4090, opened 2021-12-02
Assign courses to urls - add title parameter to options elements to show coursename & code

juan-cortizas-ponte:
Sometimes the course names are too long, this can be a problem in the tool to assign courses to urls; in case of long-named courses, the administrator will not be able to see the full name or course code.

This hack adds a title parameter to the option elements, using the course name and code as the value.

Hovering over the select elements will show course names and codes in a tooltip.

imagen

### DIFF
--- a/main/admin/access_url_edit_courses_to_url.php
+++ b/main/admin/access_url_edit_courses_to_url.php
@@ -197,7 +197,7 @@ $url_list = UrlManager::get_url_data();
                             <?php
                             foreach ($no_course_list as $no_course) {
                                 ?>
-                                <option value="<?php echo $no_course['id']; ?>"><?php echo $no_course['title'].' ('.$no_course['code'].')'; ?></option>
+                                <option value="<?php echo $no_course['id']; ?>" title="<?php echo $no_course['title'].' ('.$no_course['code'].')'; ?>"><?php echo $no_course['title'].' ('.$no_course['code'].')'; ?></option>
                             <?php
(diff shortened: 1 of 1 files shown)


main/admin/access_url_edit_courses_to_url.php
```diff
@@ -197,7 +197,7 @@ function remove_item(origin) {
                             <?php
                             foreach ($no_course_list as $no_course) {
                                 ?>
-                                <option value="<?php echo $no_course['id']; ?>"><?php echo $no_course['title'].' ('.$no_course['code'].')'; ?></option>
+                                <option value="<?php echo $no_course['id']; ?>" title="<?php echo $no_course['title'].' ('.$no_course['code'].')'; ?>"><?php echo $no_course['title'].' ('.$no_course['code'].')'; ?></option>
```
ywarnier, line 200: You need to escape this. If there is a `"` into your course title, it will break the HTML (and the whole page).
I believe using htmlentities() should work for that. To test, make sure your course title has a `"` it is and check if it shows correctly with and without htmlentities().

ywarnier (changes requested):

main/admin/access_url_edit_courses_to_url.php
juan-cortizas-ponte: Fixed with htmlentities

ywarnier (approved):

PR #4096, opened 2021-12-07
Translate HTML works with CSS - refs BT#19389

AngelFQC:


NicoDucou:
I accepted this PR but only for parkur, it should not be included in 1.11.x because this does not work with Scorm package. The content of the items do not show unless refreshing the page.
I note it here to remember why we decided not to include it in 1.11.x

Also in Master there is already a new way of managing the content translation so it will not be passed to Chamilo Master neither.

AngelFQC:
> I accepted this PR but only for parkur, it should not be included in 1.11.x because this does not work with Scorm package. The content of the items do not show unless refreshing the page. I note it here to remember why we decided not to include it in 1.11.x

Some fixes were added here https://github.com/chamilo/chamilo-lms/commit/34d45f6b3b45be991210252af8226296e3cfa0cb https://github.com/chamilo/chamilo-lms/commit/866ebedf98200dfb87d9c817cf5248c986702638

PR #4101, opened 2021-12-14
Improve loading translate html fqc bt19389

christianbeeznest:


AngelFQC:
Closed in favor of #4096

PR #4113, opened 2021-12-27
[BUG] USERMANAGER::get_user_id_of_course_admin_or_session_admin does not filter by session id

juan-cortizas-ponte:
USERMANAGER::get_user_id_of_course_admin_or_session_admin does not filter for the current session so the returned value could be wrong if a course has more than one session


ywarnier:
The current logic is that it takes the last record if there are several sessions (or teachers within that course-session), but I agree that this solution is *better*.
api_get_session_id is supposed to return 0 or some positive value, so effectively checking for empty() is enough before using that session number in the query.
There is still the possibility of having several teachers/coaches for the same course in the session, but at least we get one of them in return (the last result from the query).

PR #4132, opened 2022-01-18
Search diagnosis - Add searching by course multilanguage and count of users - refs BT#19560

christianbeeznest:


main/admin/course_edit.php
```diff
@@ -295,6 +295,17 @@
     true
 );
 
+if (api_get_configuration_value('allow_course_multiple_languages')) {
+    // Course Multiple language.
+    $languages = api_get_languages();
+    $cbMultiLanguage = $form->getElementByName('extra_multiple_language');
+    if (isset($cbMultiLanguage)) {
+        foreach ($languages['folder'] as $langFolder) {
+            $cbMultiLanguage->addOption($langFolder, $langFolder);
```
NicoDucou, line 304: We can see in the selector de language folder but we should see the translation, the folder name should be the value that is sent by the form but the user should see the translation in its language (there should be a get_lang[$langFolder] or algo asi para uno de los parametros

christianbeeznest: Thanks @NicoDucou , the fix is add in the next commit.

PR #4155, opened 2022-02-03
Template: Add video_context_menu_hidden configuration setting

AngelFQC:
Hide the context menu on video player


ywarnier:
Looks good. The configuration setting name is ok too.

PR #4202, opened 2022-03-10
Api platform: Add apis for extra fields - refs BT#19740

christianbeeznest:


ywarnier:
The code looks good to me. Couldn't test, but we'll have time for that in the overall quality checks for 2.0.

PR #4208, opened 2022-03-16
Reporting: Add new reports from MJTecnoid - refs BT#19806

christianbeeznest:


NicoDucou:
I've added the translation needed with this commit :  https://github.com/chamilo/chamilo-lms/commit/6f6b999033709d57c1a10224c5c7329812b9a90c

PR #4224, opened 2022-03-25
Fix the remind course expiration cron

fabiowoj:
The WHERE conditions has a wrong variable. Just change $sessionid$userids to $sessionid


ywarnier:
Actually no, this is correct as it is (however bad the choice of name for this variable is). As you can see on line 69, $userIds is in fact a query part (" AND id_user NOT IN ..."). This is only needed if you want to exclude some users, but it is correct.

PR #4248, opened 2022-04-19
Editor : Fix tinymce not defined loaded inside a lp - refs BT#19231

christianbeeznest:


ywarnier:
LP creation broken for now. Cannot test.

ywarnier:
Checked, tested and approved.

PR #4254, opened 2022-04-22
Session : Fix error adding template to create a session - refs BT#19231

christianbeeznest:


ywarnier:
It is better now than previously.
There's a small conflict on a filtering operation (not sure why as your filtering seems fine - leave the conflict, I will resolve it during merge).
But I think the new session should also inherit the courses and coaches (not the students) and the visibility from the template session and I don't see that with your PR.

ywarnier:
@christianbeeznest still conflicting

ywarnier:
This issue has been solved in the meantime and the PR doesn't bring anything new. Sorry for the delay!

PR #4259, opened 2022-04-25
Session List: Courses report export - refs BT#19878

lautarobeeznest:
I used the variable "$Export2ZIP" because I thought it was the most convenient for the button.

### DIFF
--- a/main/session/session_list.php
+++ b/main/session/session_list.php
@@ -116,10 +116,12 @@ switch ($action) {
         $zip = new PclZip($tempZipFile);
         $csvList = [];
 
-        foreach ($sessionList as $sessionId) {
+        foreach ($sessionList as $session_id) {
             $em = Database::getManager();
             $sessionRepository = $em->getRepository('ChamiloCoreBundle:Session');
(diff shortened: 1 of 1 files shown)


main/session/session_list.php
```diff
@@ -108,6 +110,137 @@
         header('Location: '.$url);
         exit;
         break;
+    case 'export_multiple':
+        $sessionList = explode(',', $idMultiple);
+        $tempZipFile = api_get_path(SYS_ARCHIVE_PATH).api_get_unique_id().'.zip';
+        $zip = new PclZip($tempZipFile);
+        $csvList = [];
+
+        foreach ($sessionList as $session_id) {
```
ywarnier, line 119: Use `$sessionItemId`, then define `$session_id = $sessionItemId;` right before the call to `get_user_data()` (which uses `$session_id` as a global variable)

ywarnier (changes requested):

NicoDucou:
At a functional level it works fine, so once the code revision is good it can be merge for me.

ywarnier (approved):

PR #4261, opened 2022-04-25
Session - Fix gradebook copied from main course when session is created - refs CT#3960

christianbeeznest:


ywarnier:
Perfect PR.

PR #4274, opened 2022-05-06
Work - Fix next button in pagination of works for trainers - refs BT#19877

christianbeeznest:


NicoDucou:
@lautarobeeznest the problem was fixed with this commit so you can have a look at where the issue was.

PR #4323, opened 2022-06-14
Structure db change - Replace *_session_id fields with session_id in track_e_* tables - refs CT#4318

christianbeeznest:


src/CoreBundle/Entity/TrackEAccess.php
```diff
@@ -148,25 +148,25 @@ public function getAccessTool()
...
-     * Set accessSessionId.
+     * Set sessionId.
      *
      * @return TrackEAccess
      */
-    public function setAccessSessionId(int $accessSessionId)
+    public function setSessionId(int $sessionId)
     {
-        $this->accessSessionId = $accessSessionId;
+        $this->accessSessionId = $sessionId;
```
ywarnier, line 157: I think this should be `$this->sessionId = ...`

ywarnier (changes requested):
One issue to fix

Issue #3279, opened 2020-06-05
Survey: Removing the last option in a question does not save the question title (minor)


ywarnier:
- Create a survey
- Create a question (second type)
- Create 3 options
- Save
- Edit the question
- Change the question's text
- Change the first option's text
- Remove the last option (using the "Remove" button)
- The first option's change has been saved, but not the question title's change

ywarnier:
Still valid issue, although now none of the buttons to remove an answer actually work.

christianbeeznest:
Corregido en este PR https://github.com/chamilo/chamilo-lms/pull/4329

ywarnier:
Checked and confirmed this is fixed. Thanks!

PR #4329, opened 2022-06-17
Survey - Fix Removing the last option in a question - refs #3279

christianbeeznest:


ywarnier:
I cannot test this PR because when I try to create a 3rd option, as described in issue #3279, the form submits and goes back to the list of questions. It does not add or edit the question nor the possible answers.

christianbeeznest:
I close this PR to create other with last changes in master.

PR #4340, opened 2022-07-01
fix password encryption conversion

renaudlemaire:
Wasn't working, so I fixed it.

Pass detection return bad type (bool instead string)
Users couldn't log in if password conversion is activated and detected encryption is the same as configured encryption. 
If password encryption was different, new encryption wasn't made.


ywarnier:
Included with a minor code style change (as reported by CSfixer)

PR #4360, opened 2022-07-28
Quiz: Add new question type Multiple Answer Dropdown - refs BT#20086

AngelFQC:


main/exercise/MultipleAnswerDropdown.php
```diff
@@ -0,0 +1,155 @@
...
+    {
+        global $text;
+
+        parent::createForm($form, $exercise);
+
+        $objExe = ChamiloSession::read('objExercise');
+
+        $form->addTextarea(
+            'list_text',
+            [get_lang('AnswerList'), get_lang('EnterListOfAnswersOneAnswerByLine)')],
```
NicoDucou, line 27: It seems there is an extra ) at the end of the variable name that should not be there.

PR #4398, opened 2022-09-26
Plugin: Azure + OAuth: Redirect after not allow login form

AngelFQC:


ywarnier:
This must be combined with https://github.com/chamilo/chamilo-lms/commit/2b5fe974fb29842bb77393b3e7a631afc12c4df5 to work for resources in courses, but otherwise this PR is tested, reviewed and approved.

Issue #2069, opened 2017-07-10
require errors being generated when displaying SCORM package in 'embedded' mode


moloko:
As you'll have seen from #2062 and #2063 I've been doing some testing of the [Adapt Learning Framework](https://github.com/adaptlearning/adapt_framework) in Chamilo

Whilst testing, I noticed that if I set the course to launch in 'embedded' view mode (the default), it doesn't display and require.js is generating an error in the console:
![chamilp-require-error-firefox](https://user-images.githubusercontent.com/1249597/28036319-bb3f334c-65af-11e7-945d-d6068f3d9243.png)

We've done a bit of digging and it seems that Chamilo is injecting some <script> tags into our content
![chamilo-script-injection](https://user-images.githubusercontent.com/1249597/28036394-05bf89a8-65b0-11e7-979b-7451d217d53d.png)

Specifically the problem seems to be originating from https://1.11.chamilo.org/main/glossary/glossary.js.php which uses jquery-ui.min.js - which has a define statement with a missing name - and  because Adapt uses requirejs, the module is trying to register with Adapt's requirejs library but with no name, which isn't allowed - hence the error.

If you want to see this for yourself go to the aabbb course on 1.11.chamilo, select learning paths > Adapt Version 2.0 demonstration.

Is there a reason for injecting stuff into the course content? This seems to us to be a bit of a risky thing to do as it can cause problems exactly like this.

I also got similar problems using all other view modes except fullscreen.

jmontoyaa:
Yes, Chamilo injects those js and also loads the glossary js .php file ... I think we need an option in the LP in order to don't pre load those files.

ywarnier:
Let's add this option to v2.0 (scheduled for 2018).

sinozzuke:
Hi, I'm trying to use scorms from Adapt Learning and can reproduce the problems @moloko had. As the release of Chamilo 2.0 didn't happen yet, is there a workaround so we can patch this on our Chamilo 1.11.8 installation and be able of using these scorms?

sinozzuke:
Some advances in my investigations:
Doing some research I have used campus.chamilo.org for uploading a SCORM test file, and it works like a charm. So, I have emailed them to know what version of chamilo are using but has not answer yet. 

[build.zip](https://github.com/chamilo/chamilo-lms/files/2827758/build.zip)

christianbeeznest:
[...]

PR #4403, opened 2022-09-29
Lp - Fix loading ressources inside scorm iframe - refs CT#20060

christianbeeznest:


ywarnier:
Reviewed and approved but not tested. See #2069

PR #4430, opened 2022-10-20
Document - Fix rename files to check correct document path on disk - refs BT#20137

christianbeeznest:


ywarnier:
Looks good to me. Waiting for validation by @NicoDucou in BT#20137

ywarnier:
Not sure where `move()` comes from. `rename()` is a known PHP function https://www.php.net/rename but not `move()`.

NicoDucou:
> Not sure where `move()` comes from. `rename()` is a known PHP function https://www.php.net/rename but not `move()`.

The move function is in main/inc/lib/fileManage.lib.php of Chamilo. I thought the problem was due to an access to the filesystem of the server that did not work with the rename function and was the problem that we just corrected in Chamilo, but this was not the case. The problem is due to the fact that we try to rename the file to a location that do not exist (in the new path for the file there are folders that do not exist on the server).
I'm closing this PR because it is not usefull finally.

PR #4440, opened 2022-10-27
Learnpath: Add visual warning when the time spent in a LP is less than necessary in student reports - refs GH#4299

BorjaSanchezBeezNest:


ywarnier:
Reviewed, tested and approved.

PR #4453, opened 2022-11-10
Email - Add username remember password link to the subscription to session confirmation email

juan-cortizas-ponte:
This PR enable the option to add the username of the user to the subscription to session confirmation email

The changes include a two new parameters on configuration.php, the first one to add the username to the email template:

`$_configuration['email_template_subscription_to_session_confirmation_username'] = false;`

if this setting is set to true the username value is added to the email template

The second one to add the remember password link

`$_configuration['email_template_subscription_to_session_confirmation_lost_password'] = false;`

if this setting is set to true the password link is added to the email template

imagen


ywarnier:
Hola @juan-cortizas-ponte 
Tu PR está marcado como "Draft". Lo es, o es un error al mandarlo ?

juan-cortizas-ponte:
> Hola @juan-cortizas-ponte Tu PR está marcado como "Draft". Lo es, o es un error al mandarlo ?

Despiste mío, quedó pendiente de unos cambios y lo dejé en draft, ya están realizados.

ywarnier:
OK but remember that changes to language files are not sent through PRs (you document them here and we set them in the translation application, then import from there).
I could just change this manually myself in the translation application, but this is a special case : you are modifying an existing variable with a gain-of-function (to re-use a currently popular term), meaning you are *adding* an optional text to the translation. This, in turn, means that other languages users will not be aware that this feature is available unless they find (probably randomly) that the language variable must be changed in the translation application.

To avoid this, our procedure is to create a new language variable, event if 90% of the text is the same. This way, translators realize there's something new to be translated.

I will be managing this in this case, but remember that language variables have very specific update procedures.

ywarnier:
Language variables that include a '%s' must be named with an "X" so I will rename those variables.

juan-cortizas-ponte:
> OK but remember that changes to language files are not sent through PRs (you document them here and we set them in the translation application, then import from there). [comment removed as not applicable to this PR - mistake on my part] I will be managing this in this case, but remember that language variables have very specific update procedures.

OK, thanks for the advice, I will take it into account to not repeat this mistake

PR #4455, opened 2022-11-11
Announcement - Fix url portal using multiple urls in announcement mails - refs BT#20340

christianbeeznest:


main/inc/lib/message.lib.php
```diff
@@ -892,7 +899,11 @@ public static function send_message_simple(
             $sender_id,
             $directMessage,
             0,
-            $smsParameters
+            $smsParameters,
+            false,
```
NicoDucou, line 903: There is a paramater missing here. There should be 2 false

main/inc/lib/notification.lib.php
```diff
@@ -384,6 +391,18 @@ public function formatContent($messageId, $content, $senderInfo)
             }
         }
 
+        $accessConfig = [];
+        $useMultipleUrl = api_get_configuration_value('multiple_access_urls');
+        if ($useMultipleUrl && $checkUrls) {
+            $accessUrls = api_get_access_url_from_user($senderInfo['user_info']['user_id']);
```
NicoDucou, line 397: It should be $senderInfo['user_id'] instead of $senderInfo['user_info']['user_id'] because $senderInfo['user_info'] does not exist

main/inc/lib/AnnouncementEmail.php
```diff
@@ -323,15 +323,15 @@ public function sendAnnouncementEmailToMySelf()
...
+     * @param bool $checkUrls            It checks access url of user when multiple_access_urls = true
      *
      * @return array
      */
     public function send($sendToUsersInSession = false, $sendToDrhUsers = false, $senderId = 0, $directMessage = false, $checkUrls = false)
     {
         $senderId = empty($senderId) ? api_get_user_id() : (int) $senderId;
         $subject = $this->subject($directMessage);
-
+        $courseId = $this->course('id');
```
NicoDucou, line 334: It should be $courseId = $this->course['real_id'];

PR #4474, opened 2022-11-29
Exercise: Fix hotspot image display when finish or see the result of the exercise - refs GH#4315

BorjaSanchezBeezNest:


public/main/exercise/hotspot_answers.as.php
```diff
@@ -168,13 +168,13 @@
     $qb
         ->select('a')
         ->from(CQuizAnswer::class, 'a')
-        ->where($qb->expr()->eq('a.cId', $courseId));
+        ->where($qb->expr()->eq('a.iid', $courseId));
```
ywarnier, line 171: This pre-filter does not apply anymore as c_id is gone.

ywarnier (changes requested):

public/main/exercise/hotspot_answers.as.php
ywarnier: You can remove this line instead of changing it to iid

PR #4494, opened 2022-12-13
Agenda: Allow import course events in CSV - refs BT#20356

AngelFQC:
Refs BT#20356

### DIFF
--- a/main/admin/index.php
+++ b/main/admin/index.php
@@ -396,6 +396,15 @@ if (api_is_platform_admin()) {
         'url' => api_get_path(WEB_CODE_PATH).'calendar/agenda_js.php?type=admin',
         'label' => get_lang('GlobalAgenda'),
     ];
+
+    if (true === api_get_configuration_value('agenda_reminders')) {
+        $items[] = [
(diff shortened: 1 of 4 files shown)


main/admin/index.php
```diff
@@ -396,6 +396,13 @@
         'url' => api_get_path(WEB_CODE_PATH).'calendar/agenda_js.php?type=admin',
         'label' => get_lang('GlobalAgenda'),
     ];
+
+    $items[] = [
+        'class' => 'item-agenda-reminders',
+        'url' => api_get_path(WEB_CODE_PATH).'admin/import_course_agenda_reminders.php',
+        'label' => get_lang('AgendaReminders'),
+    ];
+
```
ywarnier, line 405: Shouldn't this link appear *only* if api_get_configuration_value('agenda_reminders') is true ?

ywarnier (changes requested):
Please review the comment on showing the link even if the feature is not enabled. I think it introduces confusion.

ywarnier (approved):

PR #4497, opened 2022-12-14
System: Fix Behat scenario 'Enter to public password-protected course' - refs GH#1678

BorjaSanchezBeezNest:


ywarnier:
The behat run by Github still returns an error on that precise test: https://github.com/chamilo/chamilo-lms/actions/runs/3697991616/jobs/6382177704

ywarnier:
Maybe a click must be sent to #collapse-course-access-settings to make sure the setting appears ?
Maybe behat has to be updated in composer.json to find a hidden form element ?
Not sure...

BorjaSanchezBeezNest:
After several tests, it has been fixed in local, but it continues to fail in GitHub tests
The error:
`Form field with id|name|label|value|placeholder "course_registration_password" not found.`

The following has been tried:

- Upgrade Chrome driver version from '88.0.4324.96' to '108.0.5359.71'
- Upgrade Selenium driver to '4.7.2', but any test work with this version
- Change the test screen resolution from 1280x1024 to 1920x1080
- Put ID in the link that is not recognized, with the same result

ywarnier:
This PR is intentionally left "on hold" as it doesn't solve the issue in all contexts, but provide hints at a future solution.

PR #4518, opened 2022-12-29
Exercise - Add report of exercise attempts with questions score - refs BT#20366

christianbeeznest:


ywarnier:
Looks OK to me (technically speaking). Although I think the double loop in getTrackExerciseAttemptsTable() will be an efficiency issue with several thousand users and dozens of questions.

PR #4520, opened 2023-01-03
BUG - Setting a default view for the admin session list on one site of a multi campus environment could break the list on the other sites

juan-cortizas-ponte:
Chamilo has a configuration setting to set the default view for the admin session list, there are four possible values: https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/install/configuration.dist.php

```
// Set the default tab in the admin session list. Values: all, close, active, custom.
//$_configuration['default_session_list_view'] = 'all';
```

At a multicampus enviroment, if we set it for a campus, for example:

`$_configuration['default_session_list_view'][2] = 'active';`

... works on that site, but this config  breaks the list on the others sites because [SesionManager.getDefaultSessionTab()](https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/inc/lib/sessionmanager.lib.php#L9568-L9578) checks only if api_get_configuration_value('default_session_list_view') is empty:

```
    public static function getDefaultSessionTab()
    {
        $default = 'all';
        $view = api_get_configuration_value('default_session_list_view');

        if (!empty($view) ) {
            $default = $view;
        }

        return $default;
    }
```

imagen

This PR fixes it checking on SesionManager.getDefaultSessionTab() if the view name is one of the four possible values

```
    public static function getDefaultSessionTab()
    {
        $default = 'all';
        $view = api_get_configuration_value('default_session_list_view');

        if (!empty($view) && ( $view == 'all' || $view == 'close' || $view == 'active' || $view == 'custom' )) {
            $default = $view;
        }

        return $default;
    }
```

### DIFF
--- a/main/inc/lib/api.lib.php
+++ b/main/inc/lib/api.lib.php
@@ -9066,6 +9066,9 @@ function api_get_configuration_value($variable)
                 // Try to found element with id = 1 (master portal)
                 if (array_key_exists(1, $_configuration[$variable])) {
                     return $_configuration[$variable][1];
+                } else {
+                    // The value was there for other URLs but not the main URL nor the current URL
+                    return null;
                 }
             }
         }
--- a/main/inc/lib/sessionmanager.lib.php
+++ b/main/inc/lib/sessionmanager.lib.php
@@ -9570,7 +9570,7 @@ class SessionManager
         $default = 'all';
         $view = api_get_configuration_value('default_session_list_view');
 
-        if (!empty($view)) {
+        if (!empty($view) && ( $view == 'all' || $view == 'close' || $view == 'active' || $view == 'custom' )) {
             $default = $view;
         }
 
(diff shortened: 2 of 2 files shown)


ywarnier:
@juan-cortizas-ponte There's a little spacing issue in the CS fixer analysis. Could you send a patch to your PR?

ywarnier:
I'm not sure why, but this PR breaks all the behat tests (or almost all of them). I've been looking at it but cannot find a reason for that.

ywarnier:
Analysing this a bit further, it looks like you have found a whole in the api_get_configuration_value() function, but you are trying to fix it in a way that only patches your case.
A better fix would be to change this:
https://github.com/chamilo/chamilo-lms/blob/1.11.x/main/inc/lib/api.lib.php#L9068-L9070

To this:
```
                // Try to find element with id = 1 (master portal)
                if (array_key_exists(1, $_configuration[$variable])) {
                    return $_configuration[$variable][1];
                } else {
                    // The value was there for other URLs but not the main URL nor the current URL
                    return null;
                }
```

Could you change this and see if it works in your case (without the code of your PR) ? (considering you already have a good setup to test this...)

juan-cortizas-ponte:
> 
@ywarnier works fine with your fix!, thanks for the suggestion!

PR #4522, opened 2023-01-04
Calendar - Reduce field size of personal_agenda.color and c_calendar_event.color to varchar(20) - refs CT#4514

christianbeeznest:


ywarnier:
Code reviewed. Not tested.

PR #4530, opened 2023-01-10
LP: Add call HookLearningPathCreated when creating scorm

AngelFQC:
Refs BT#20443

### DIFF
--- a/main/lp/learnpath.class.php
+++ b/main/lp/learnpath.class.php
@@ -759,6 +759,7 @@ class learnpath
             $courseInfo = api_get_course_info();
         }
 
+        $em = Database::getManager();
         $tbl_lp = Database::get_course_table(TABLE_LP_MAIN);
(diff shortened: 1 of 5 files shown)


main/lp/learnpath.class.php
```diff
@@ -777,7 +778,13 @@ public static function add_lp(
...
             $expired_on = null;
         } else {
-            $expired_on = Database::escape_string(api_get_utc_datetime($expired_on));
+            $expired_on = api_get_utc_datetime($expired_on, false, true);
+        }
+
+        if (empty($publicated_on)) {
+            $publicated_on = null;
+        } else {
+            $publicated_on = api_get_utc_datetime($publicated_on, false, true);
```
AngelFQC, line 787: This condition doubles the time difference. This should replace the previous condition.

Fixed here 9205b9c

PR #4548, opened 2023-01-18
Exercise: Fix copy exercise function

BorjaSanchezBeezNest:


ywarnier:
Looks good. Code Climate will complain about PHPDoc lines not ending with a period (`.`), but I can live with that.
There seems to be a conflict in validation between format code and codeclimate, whereby it requires some spaces around an assignation `=` sign and the other requires no space, but I can also live with that.

PR #4557, opened 2023-01-23
Course copy: Fix calendar events don't copy from backup - refs BT#20503

christianbeeznest:


NicoDucou:
Todo conforme a nivel funcional

PR #4565, opened 2023-01-24
Learnpath: Rename c_lp.publicated_on to c_lp.published_on -refs CT#4523

christianbeeznest:


ywarnier:
Code looks good. Didn't test it yet but we still have time for that.

PR #4567, opened 2023-01-25
Course copy: Rename coursecopy folder into course_copy - refs CT#4535

christianbeeznest:


ywarnier:
Reviewed, tested and approved.

PR #4568, opened 2023-01-25
Reporting: Rename mySpace folder to my_space - refs CT#4536

christianbeeznest:


ywarnier:
Reviewed, tested and approved. Could not find any issue due to the change.

PR #4586, opened 2023-02-03
Gradebook: Add gradebook_display_extra_stats config setting

AngelFQC:
[1] = Ranking
[2] = Best Score
[3] = Average

Enable specific columns in gradebook table.

It deprecates the gradebook_enable_best_score config setting


main/install/configuration.dist.php
```diff
@@ -1088,8 +1088,16 @@
 // - launch composer install to rebuild the autoload.php
 //$_configuration['allow_portfolio_tool'] = false;
 
+// DEPRECATED: Use gradebook_display_extra_stats instead.
```
ywarnier, line 1091: `// DEPRECATED: gradebook_enable_best_score is deprecated. Use gradebook_display_extra_stats instead.`

ywarnier (changes requested):
Just a minor documentation change to make it impossible to confuse.

PR #4587, opened 2023-02-04
Admin: fix configuration variable recuperation for multiURL platform…

NicoDucou:
…s also -refs BT#20531

### DIFF
--- a/main/inc/lib/api.lib.php
+++ b/main/inc/lib/api.lib.php
@@ -9084,6 +9084,20 @@ function api_get_configuration_value($variable)
     if (isset($_configuration[$variable])) {
         if (is_array($_configuration[$variable]) && api_is_multiple_url_enabled() && is_int(array_keys($_configuration[$variable])[0])) {
             // It has been configured for at least one sub URL so we will not return the complete variable
+            /*
+	     * The idea is that if the first level key of the configuration variable is an int
+	     * then it is a multiURL configuration and if it's a string then it's a configuration that is not multiURL.
+	     * For example if in app/config/configuration.php you have set :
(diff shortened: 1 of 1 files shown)


main/inc/lib/api.lib.php
```diff
@@ -9082,7 +9082,8 @@ function api_get_configuration_value($variable)
 
     // Check if variable exists
     if (isset($_configuration[$variable])) {
-        if (is_array($_configuration[$variable]) && api_is_multiple_url_enabled()) {
+        if (is_array($_configuration[$variable]) && api_is_multiple_url_enabled() && is_int(array_keys($_configuration[$variable])[0])) {
```
ywarnier, line 9085: I don't understand the need for the `is_int()` to enter the `if`... I would understand checking if it is `!empty()`, but using `is_int()` suggests that you will ignore all settings of which the value is a string, which sounds weird...

NicoDucou: The idea is that if the configuration variable is an int entonces it is a multiURL configuration and if it's a string then it's a configuration that is not multiURL. For example if in app/config/configuration.php you have set :

```
$_configuration['ticket_project_user_roles'] = [
    'permissions' => [
        1 => [17] // project_id = 1, STUDENT_BOSS = 17
    ]
];
```
You do not want to enter in this bloc even if multiURL is activated because the option is configured globaly and you want to return the full array.

The is_int is to consider only the option that are array and configured for multiURL which means there is an int as the first level key of the array.

I hope it helps to understand why there is this need for the is_int testing to make configuration of array options available in a multiURL context.

PR #4599, opened 2023-02-08
Display course image in header, course related emails and pdf files

TheTomcat14:
The changes to the software do the following : 

- Add the possibility to upload and crop an extra course logo with the same settings as the platform logo.
- Add a setting to the Administration > Course settings that lets you decide to use the course logo (if present) instead of the platform logo
  - If this setting is turned to true, the logo will be used in the header and pdf exports and course related emails
- In order to imitate the requirements for the platform logo dimensions, the cropper was updated with the possibility to specify a min- and max ratio


ywarnier:
Hi @TheTomcat14 

Thanks for the contribution. I had a quick look at the code and it looks great.

Just a note: we do not accept database changes between minor versions, so we have a procedure which might seem counterintuitive but that is the procedure to avoid having other issues: database changes are added to main/install/configuration.dist.php as information next to the setting that has to be enabled manually in the app/config/configuration.php file (you can see a series of examples if you look for ALTER TABLE or INSERT in uppercase in main/install/configuration.dist.php).

Considering you probably didn't know that and the rest of your code is neat, we will probably update that ourselves in your PR before including it.
In the end, it will be more complex to enable the feature than what your PR was aiming for, but introducing automated database changes between minor versions would create a mess in our upgrade procedures.

Also, please be aware that the 1.11.x branch is now unsynchronized with the master branch. The master branch will be the source of the 2.0 version we intent on releasing this year. This means your added feature will not be included in 2.0 unless you submit another PR on the master branch (this can be done at a later stage but we are aiming for a beta version by the end of April, and this would mean a database freeze, so you might want to get that in before that date, or it will only appear in a future version).

ywarnier:
This has been merged (thanks @TheTomcat14 ) but with considerable changes to make it match both naming and coding conventions.
The setting (on/off) is not added to the database, so it will not appear in the Chamilo setting, but will have to be added manually to app/config/configuration.php as `$_configuration['mail_header_from_custom_course_logo'] = true;`.
The name of the uploaded file has changed, because in Chamilo it was a historical mistake to call the file "85x85", considering it's not that size at all. Now it just says "cropped" for your file when it is in cropped format.

I'm sorry this means you will have to change a few things on the portal you enabled this in, but otherwise it would have made things worst from a maintenance point of view.

You will thus have to look for the files with a little script and rename them, then add the setting to configuration.php and remove it from the settings_current table where it is right now but will not be used by the code (just in your case, that is).

This was still a very good contribution, though !

PR #4607, opened 2023-02-13
Internal: Add options to allow configuring mail with XOAuth method

AngelFQC:


main/inc/lib/api.lib.php
```diff
@@ -9426,6 +9428,28 @@ function api_mail_html(
     }
 
     $mail = new PHPMailer();
+
+    if (isset($platform_email['XOAUTH2_METHOD']) && $platform_email['XOAUTH2_METHOD']) {
```
ywarnier, line 9432: Sounds like it can be simplified to `!empty($platform_email['XOAUTH2_METHOD'])`.

app/config/mail.conf.dist.php
```diff
@@ -41,3 +41,12 @@
 // showing it as a loose JSON string to the final user. If this is your case,
 // you might want to set the variable below to 'false' to disable this header.
 $platform_email['EXCLUDE_JSON'] = false;
+
```
ywarnier, line 44: Should come with some doc:
```
// Fill the following only for mail services with OAuth2.0 authentication. Otherwise leave untouched.
```

ywarnier (changes requested):
Minor adjustments required

PR #4629, opened 2023-03-03
Text2Speech Plugin

AngelFQC:


ywarnier:
After doing some testing, we are rejecting this PR (and its daddy #4623 ) because the pre-compiled TTS models available, as far as we could see, are very limited in size and generate only about 50 characters of TTS conversion, and have notable quality issues so far (in Spanish, numbers are skipped, and in French there is background noise). As the Chamilo project, we do not (sadly) have the resources to maintain (train, make available, document) numerous models (one for each language) at this point. The added value (converting text to audio to provide an additional media in the learning path or elsewhere) doesn't seem to be worth the effort. Browser plugins are available for blind users and Chamilo complies with WCAG AA for most content, making it accessible already.

This PR is more adapted to Chamilo than #4623 as it offers a way for teachers to re-generate the audio track after modifying the item text, so I'm closing #4623 and leaving #4629 open for future integration when we find better readily available trained models to use.

Savvy Chamilo admins can most probably use this PR and compile their own TTS model from the Mozilla TTS sources to obtain better results, so double reason to leave this open.

ywarnier:
Other source worth investigating: https://github.com/facebookresearch/fairseq/tree/main/examples/mms

ywarnier:
Abandonned considering the enormous progress of AI over the years since this PR was first opened.
New solutions will probably be integrated in the future, like Open Source speech generation models.

PR #4630, opened 2023-03-06
Fix imagegif() call, function has no third parameter

TheTomcat14:
The imagegif() function has no third parameter, the $compress parameter can't be used.
The function is used in GDWrapper::send_image()


ywarnier:
Thanks @TheTomcat14 !

PR #4653, opened 2023-03-21
Composer: Generate ORM proxies post composer install

AngelFQC:


ywarnier:
Abandonned in favor of #4656

PR #4658, opened 2023-03-23
Certificates: Issued certificates CSV report

juancp-contidosdixitais:
A new feature for Chamilo, a certificate export report for the certificates issued 

This new feature adds a button to allow download a CSV with the user data, certificate issue date, score and optional extra user fields

image

Optional extra user fields must be configured in configuration.php using the new configuration parameter "certificate_export_report_user_extra_fields":

$_configuration['certificate_export_report_user_extra_fields'] = ['extra_fields' => ['office_address', 'office_phone_extension']];


contidos:
In terms of users what this new feature does is to export a csv file that contains: 

id
username
name
surname
score (result on its certificate)
date in wich the certificate was generated as it appears on the interface list
extra fields suchs ad DNI number or any other that we select in configuration.php

This feature helps companies or organizations that need to report on wich new employees have been certificated in an specific range of time.

ywarnier:
@juancp-contidosdixitais Don't forget to mention when there is a new language term (ExportCertificateReport in this case) so we can add it to the translation system.

PR #4687, opened 2023-04-20
Attachment upload fails in Calendar::editEvent()

TheTomcat14:
Uploading attachments to calendar events is not possible, file processing is skipped. 
Also, new attachments need to be added instead of being updated.


ywarnier:
The creation form for events was fine but indeed, the edition wasn't.
I couldn't reproduce the issue of updating the current file when adding a new one, though (that works for me before your patch).
The PR looks good to me. I tested it and it does fix the upload on edit issue. Couldn't find any adverse effect.
Thanks for your contribution!

PR #4713, opened 2023-05-19
Fix course detail title buttons

daniboygg:
* Create BaseButton to create a consistent button across all code
* Create BaseIcon to restrict the icons used across all code
* Create BaseMenu as starting point for menu component consistent across all code
* Fix not loaded header logo of chamilo, importing in the component make the url resolution correctly
* Fix colors of tailwind according to figma
* Disable vuetify due to a crash with tailwind styles


assets/vue/main.js
```diff
@@ -285,7 +285,7 @@ app
     .use(Quasar, quasarUserOptions)
     .use(VueFlatPickr)
     //.use(VuelidatePlugin)
-    .use(vuetify)
+    // .use(vuetify)
```
daniboygg, line 288: There is a conflict between tailwindcss library and vuetify using clasess like color-primary, this is needed for tailwind to work

assets/vue/components/layout/TopbarLoggedIn.vue
```diff
@@ -52,6 +52,8 @@ import Avatar from "primevue/avatar";
 import Menu from "primevue/menu";
 import { usePlatformConfig } from "../../store/platformConfig";
 
+import headerLogoPath from "../../../../assets/css/themes/chamilo/images/header-logo.svg";
```
daniboygg, line 55: This solves the problem when using dev-server that did not show chamilo header icon

Issue #2661, opened 2018-09-11
Remove unused plugins / Fix plugins


jmontoyaa:
There are some plugins that won't work, because of the different changes in Chamilo v2

Can we update this table to know which will be deleted or fixed?

I created this table so you can edit it. This is just a proposal.

Checked in V2 | Plugin name | what to do | Folder in C2 |
| -- | ------------ | ------------- | ---- |
|X | add_cas_login_button | you can delete it, check if silex/security has one | - |
|X | add_cas_logout_button | you can delete it, check if silex/security has one | - |
|X | add_facebook_login_button | you can delete it, check if silex/security has one | - |
|X | add_shibboleth_login_button | you can delete it, check if silex/security has one | - |
| | advanced_subscription | keep and integrate later (after 2.0), so need to prepare the structure. Error when calling api_get_setting('advanced_subscription') | - |
| | ai_helper | Integrate to core (done) | - |
| | azure_active_directory | keep or replace with equivalent (this might be managed by a component now?). Assign to @AngelFQC | - |
| | bbb | keep/test/fix | Bbb |
| | before_login | keep/test/fix | BeforeLogin |
| | buycourses | keep/test/fix & (integrate option)  | BuyCourses |
| | card_game | ? | CardGame |
| | check_extra_field_author_company | ? | CheckExtraFieldAuthorCompany |
| | cleandeletedfiles | ? | CleanDeletedFiles |
|X | clockworksms | delete: Not aware of anyone using it (@ywarnier) and no effort from external (for pay) provider to contribute | - |
| | courseblock | keep/test/fix | CourseBlock |
| | coursehomenotify | ? | CourseHomeNotify |
| | courselegal | keep/test/fix & integrate | CourseLegal |
| | createdrupaluser | delete | - |
| | customcertificate | keep/test/fix & integrate (replace current certificate tool?) | CustomCertificate |
| | customfooter | keep/test/fix & integrate | CustomFooter |
| | dashboard | This is a special case allowing for more blocks. Not really a plugin. | Dashboard |
| X | date | delete: Example initial super basic plugin - not accurate anymore. | - |
| :heavy_check_mark: | dictionary | tested/works :heavy_check_mark: | Dictionary |
| | embedregistry | keep/test/fix | EmbedRegistry |
| | exercise_signature | ? | ExerciseSignature |
| | ext_auth_chamilo_logout _button_behaviour | keep/integrate | ExtAuthChamiloLogoutButtonBehaviour |
[...]

PR #4719, opened 2023-05-24
Plugins: Fix Readme dialog popup and configure option after saving - refs #2661

christianbeeznest:


ywarnier:
Not much to reject here... :sweat_smile:

PR #4724, opened 2023-05-30
Fix link to group in calendar view

TheTomcat14:
When accessing the calendar from within a course group, the group data gets lost when editing an event.
The breadcrumb does not contain the group info anymore.

### DIFF
--- a/main/calendar/agenda.php
+++ b/main/calendar/agenda.php
@@ -14,10 +14,16 @@ if (!empty($course_info)) {
 
 $action = isset($_GET['action']) ? Security::remove_XSS($_GET['action']) : null;
 
+$group_id = api_get_group_id();
+
 $url = null;
 if (empty($action)) {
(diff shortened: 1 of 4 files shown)


ywarnier:
I'm not sure this PR is complete...
When testing it, I don't see more group context being maintained than without the PR. I do see that we loose the gidReq on the edition page without your PR, but then your PR doesn't change that (the breadcrumb continues to loose the group context).
Could you explain what should be different with your PR?

ywarnier:
Sorry, got it: this was due to the JS cache. Now I see it.
Thanks for your contribution!

ywarnier:
Couldn't merge it from here. I merged it offline but couldn't attribute the change to you directly (added it in the comment of the commit): https://github.com/chamilo/chamilo-lms/commit/dbe0e52c548a5d38dee86a47df38801bdd2499c2, sorry about that.

PR #4725, opened 2023-05-30
Fix for substr() on DateTime in Wiki

TheTomcat14:
The substr() function throws an error because lastime is a DateTime instead of a string


ywarnier:
Looks good to me.
Thanks for your contribution!

Issue #3787, opened 2021-02-23
Resource order


jmontoyaa:
Being able to order a list of resources

ywarnier:
Add a `display_order` int field in the `resource_link` table to allow resources to be sorted in any specific context (for the same c_id + session_id + group_id, the display_order should be configurable).
This implies managing correctly actions like `moveUp()`, `moveDown()`, `delete()`, `add()`, etc, so they always update the `display_order` correctly.
By default, `add` should add the new resource at the end of the `display_order` (MAX(display_order) + 1), while `delete` should remove the resource and make a `moveUp()` for all the following resources, as the deleted resource has freed one display_order from the list.

christianbeeznest:
It is added field display_order in this PR https://github.com/chamilo/chamilo-lms/pull/4636

Thanks for confirmation.

ywarnier:
Hola @christianbeeznest 
Quisiera que revisaras https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/doc/sortable.md (sugerido por Angel) para hacer el ordenamiento en vez de hacerlo a mano. Ya usamos doctrine-extensions, así que no introduce ninguna dependencia adicional, y sería más sano poder basarnos en esta para el largo plazo.

christianbeeznest:
Hi @ywarnier 

It is added the sortable in this PR https://github.com/chamilo/chamilo-lms/pull/4728

Thank you.

daniboygg:
@NicoDucou sure! I can help with this. But I have a couple of questions for @christianbeeznest:

1. Right now the endpoint to retrieve the course is `../course/${courseId}/home.json` (https://github.com/daniboygg/chamilo-lms/blob/master/assets/vue/views/course/CourseHome.vue#LL220C22-L220C53). I cannot see the documentation about this webservice in https://2.chamilo.org/api/, is this the right call for retrieving this?
2. How can I change the order in the backend? I don't know where to find documentation about this feature. Is there a pull request with this feature?

christianbeeznest:
Hi @daniboygg 

1. Yes, the endpoint to retrieve the course is ../course/${courseId}/home.json . I think from course home when the tool icons are moved it should send the current positions to that path, so it could be updated in the database with some changes in backend (I could check that part).

[...]

PR #4728, opened 2023-05-31
Resource: Add doctrine Sortable to simply the process - refs #3787

christianbeeznest:


christianbeeznest:
I close this ticked due conflicts with last master version.

PR #4734, opened 2023-06-02
Record audio from document tools - refs #4731 

### LINKED ISSUE #4731: [C2] Allow users to record and upload an audio in the documents tool
There is already a button for this, but is disabled right now

image

There is a component in vue called AudioRecorder.vue https://github.com/chamilo/chamilo-lms/blob/master/assets/vue/components/AudioRecorder.vue. Reuse this component and adapt it to work with this feature (already present in chamilo 1.11)

This is a follow up of the task #4684

daniboygg:
Fix #4731


assets/vue/views/documents/DocumentsList.vue
```diff
@@ -584,4 +598,18 @@ function showUsageDialog() {
...
+
+function recordedAudioSaved() {
+  // TODO show a notification about upload success
+  isRecordAudioDialogVisible.value = false
+  onUpdateOptions(options.value)
+}
+
+function recordedAudioNotSaved() {
+  // TODO show a notification about upload failure
+}
```
daniboygg, line 614: To show the notifications we need to await to the pull request https://github.com/chamilo/chamilo-lms/pull/4729to be merged since the notification composable is made in it

PR #4737, opened 2023-06-06
Update agenda.lib.php

LudiscapeXApi:
Add function deleteEventIfAlreadyExit
Prevent duplicate event to the course calendar when import large outlook calendar


ywarnier:
I have included it with the changes in the additional commit (method name change + addition of sessionId). Please test it to confirm that it works with the sessionId as well (you can easily reproduce the use case it's supposed to fix)

ywarnier:
@LudiscapeXApi

PR #4739, opened 2023-06-07
Plugin: OAuth2: Allow to set values for user status from oauth2 response

AngelFQC:
refs BT#20784

### DIFF
--- a/plugin/oauth2/src/OAuth2.php
+++ b/plugin/oauth2/src/OAuth2.php
@@ -382,7 +382,7 @@ class OAuth2 extends Plugin
         return $fieldValue->getCreatedAt() >= $lastLogin->getLoginDate();
     }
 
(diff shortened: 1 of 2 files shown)


plugin/oauth2/src/OAuth2.php
```diff
@@ -445,12 +470,9 @@ private function updateUser($userId, $response)
                 $user->getEmail()
             )
         );
-        $user->setStatus(
-            $this->getValueByKey(
-                $response,
-                $this->get(self::SETTING_RESPONSE_RESOURCE_OWNER_STATUS),
-                $user->getStatus()
-            )
+        $status = $this->mapUserStatusFromResponse(
```
NicoDucou, line 473: No seteas el status.
No faltaria $user->setStatus($status); ?

PR #4755, opened 2023-06-21
Fix audio display for document tool - refs #4754

### LINKED ISSUE #4754: [C2] Show icon and correct player for audio files
When a user click on a file the controls are broken. A black bar show but no controls to play and pause the audio are shown. In addition, the icon of the audio file is a generic file icon.

daniboygg:
Fix #4754

### DIFF
--- a/assets/vue/components/documents/DocumentEntry.vue
+++ b/assets/vue/components/documents/DocumentEntry.vue
@@ -39,7 +39,7 @@ const props = defineProps({
 })
 
 const cidQuery = useCidReq()
-const {isFile: utilsIsFile, isImage, isVideo} = useFileUtils()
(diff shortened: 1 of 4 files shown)


assets/vue/components/documents/DocumentEntry.vue
```diff
@@ -52,6 +52,9 @@ const dataType = computed(() => {
   if (isVideo(props.data)) {
     return 'video'
   }
+  if (isAudio(props.data)) {
+    return 'video'
+  }
```
daniboygg, line 57: This property is set in :data-type="dataType". I couldn't find anything in documentation about audio. I couldn't use iframe and set the height of controls either. This approach works but show controls like a video. 

It's not ideal but at least user can see the audio controls.

PR #4761, opened 2023-06-22
Import user tool: Results may appear incorrect when importing more than one user

juancp-contidosdixitais:
We detected an error on the user import result view.

When importing more than one user through the bulk import tool (/main/admin/user_import.php), upon completion of the process, the last record may appear with an incorrect status or directly as a duplicate of the first one. The problem lies in reusing the variable $user in the iterations of the arrays $userSaved, $userError, and $userWarning within the save_data function in main/admin/user_import.php.

As a quick and simple solution, I have chosen to make the $user in the foreach loops an assignment by reference. I don't see any inconvenience in doing so.

As a side note unrelated to the solution of this pull request, I think it might be the best solution, but I have doubts about whether it would be better to rename the variables, both the arrays and the assignments. For example:

> Array $userError to $usersError
> Asignación $user to $userError
>
> Code:
> `foreach ($usersError as $userError) { ....`

### DIFF
--- a/main/admin/user_import.php
+++ b/main/admin/user_import.php
@@ -400,7 +400,7 @@ function save_data(
             $csv_content = [];
             $csv_row = $header;
             $csv_content[] = $csv_row;
-            foreach ($userSaved as $user) {
+            foreach ($userSaved as &$user) {
                 $csv_row = [];
                 $csv_row[] = isset($user['id']) ? $user['id'] : '';
                 $csv_row[] = isset($user['FirstName']) ? $user['FirstName'] : '';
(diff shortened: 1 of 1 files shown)


ywarnier:
Hi @juancp-contidosdixitais 
Could you provide an anonymized version of a CSV file that would allow me to reproduce the issue?
Looking at your description and the code, I don't see how the `$user` variable would be an issue, but I *do* see that we are using `$user` for many different things and that it would probably be a better idea to just assign a few different variables, rather than using it by reference (`&$user`), which might have other unintended consequences.

ywarnier:
I rewrote your PR.
I now understand why you did it that way, but I felt it introduced more confusion, as the Saved/Error/Warning loops were really not using the same "$user" variable as the previous loop. So I just renamed the variable to $userItem to make sure no overlapping occured, and as far as I could verify, it fixes the issue you were reporting.
I added you as Co-Author in the commit message. Thanks for your contribution.

PR #4808, opened 2023-07-13
Plugin: Keycloak: Align properly the Keycloak login button

ivosh:
When `Keycloak` plugin is enabled, a new button for logging in via Keycloak is displayed below the normal "Login" button.
This PR fixes the visual appearance of this "Keycloak" button.

Before:
image

After:
image


ywarnier:
Thanks @ivosh

PR #4810, opened 2023-07-14
Migrations: Check files of migrations - refs BT#20846

christianbeeznest:


christianbeeznest:
I close this PR to avoid conflict, there is other PR with the fixes.

PR #4816, opened 2023-07-17
Scheduled Announcements: Add user extra fields as tags

juancp-contidosdixitais:
This PR adds a new funcionality that allows to add selected user extra fields to the tags list of the scheduled announcements.

We have to add a new configuration.php parameter, "scheduled_announcements_user_extra_fields_as_tags" to set the user extra fields to show on the editor

### DIFF
--- a/main/inc/lib/ScheduledAnnouncement.php
+++ b/main/inc/lib/ScheduledAnnouncement.php
@@ -412,6 +412,16 @@ class ScheduledAnnouncement extends Model
                                 '((lp_progress))' => $progress,
                             ];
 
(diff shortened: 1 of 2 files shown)


ywarnier:
In the certificates templating process, we include all user tags that are marked as "Visible to self" and "Can be used as filter".
I would prefer if we could use the same method here, but I'm not sure that works with your use case...

ywarnier:
I'm closing this one but I sent the changes necessary (without configuration option) to add user extra fields to the scheduled announcements, using the same method as in the certificates.

PR #4837, opened 2023-08-14
Migrations: Add link to redirect home after registration and logged - refs BT#20846

christianbeeznest:


christianbeeznest:
This PR is not necessary ahora, other redirection is already used.

PR #4865, opened 2023-09-05
dropbox tool : "view" parameter check does not display error

TheTomcat14:
The check on a valid "view" parameter does not work as expected.

When the $view variable is compared to 'sent' the operator is wrong, a single "=" is used to assign the value "sent" to the variable instead of a double "==" for comparison.

Fixing this bug revealed another bug, because in the following redirect to an error message, the $view parameter is passed on, which causes an infinite redirect loop.

As a fix the user is redirect to the "reveived" view, with an error message, because it is treated as the default view for the tool.

### DIFF
--- a/main/dropbox/dropbox_init.inc.php
+++ b/main/dropbox/dropbox_init.inc.php
@@ -283,10 +283,10 @@ $(function () {
 $checked_files = false;
 if (!$view || $view == 'received') {
     $part = 'received';
-} elseif ($view = 'sent') {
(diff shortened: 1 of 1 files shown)


ywarnier:
Checked and approved. Thanks!

ywarnier:
The change from `$view` to `received` certainly seems to be limiting for future other actions, but there's no plan to extend for now and I can't see how to break it unvoluntarily, so I'm merging this.

PR #4866, opened 2023-09-05
Survey tool - dates incorrectly formatted when editing meeting

TheTomcat14:
The start- and end-dates are incorrectly formatted when editing a doodle type survey.


ywarnier:
Checked and approved. Thanks!

PR #4879, opened 2023-09-12
Survey multiplechoiceother fix

TheTomcat14:
Export-functions of multiplechoiceother type questions are not correct: 

- the "other" option was not selected when viewing the filled-in form
- the count of other answers in the overall report was incorrect (it always displayed 1)
- the "other" text was never exported in csv of excel formats


ywarnier:
Checked and approved. Thanks!

PR #4884, opened 2023-09-18
Update fileUpload.lib.php

LudiscapeXApi:
strtolower only if the file is a .htaccess 
yverhenne https://forum.chamilo.org/viewtopic.php?f=14&t=12019&p=39874#p39874


AngelFQC:
Fixed here f680d7dec1c01f1899886ec1929ba3a1b54e2d84

PR #4885, opened 2023-09-18
Fix "deny from all" statements in .htaccess files

TheTomcat14:
"Deny from all" statements do not work in Apache 2.4. They are replaced by "Require all denied".

A check was added to make the .htaccess files work with both Apache 2.2 & 2.4.


ywarnier:
Cool! Thanks! Merged.

PR #4893, opened 2023-09-20
Session: Fix minor warnings in Chamilo installation process

BorjaSanchezBeezNest:


NicoDucou:
I did not reproduce the warning but there is nothing wrong with this PR so I accept it.

PR #4911, opened 2023-10-02
Allow storing long texts in extra field value.

TheTomcat14:
Currently the value for an extra field is stored in a varchar(255) field in the extra_field_value table of the database. 
This is to small to store for example one of the standard templates available in the WYSIWYG editor.

Would it not be better to store the value in a longtext field?


TheTomcat14:
In order to change the field in the database the following SQL needs te be executed.

ALTER TABLE `extra_field_values` CHANGE COLUMN `value` `value` LONGTEXT NULL AFTER `field_id`;

ywarnier:
Thank you for your contribution @TheTomcat14 :+1:

PR #4968, opened 2023-11-01
Language: Add sublanguage using gettext files and get_lang - refs BT#20922

christianbeeznest:


christianbeeznest:
I close this PR to avoid conflicts with last version in master

PR #5019, opened 2023-12-13
Session: Check cross-site scripting - refs BT#21289

christianbeeznest:


ywarnier:
Some of these filters were not strictly necessary, but they're not a big issue to leave active.

PR #5026, opened 2023-12-22
Update justification.php

LudiscapeXApi:
Ajout d'une icone Pièce non chargée Support #21052

### DIFF
--- a/main/auth/justification.php
+++ b/main/auth/justification.php
@@ -191,7 +191,6 @@ $(function(){
         var colorG = $(this).css("color");
         var lgtxt = $(this).text().replace(/ /g,"").length;
         if (colorG!="green"&&colorG!="rgb(0, 128, 0)"&&lgtxt>3) {
-            alert(lgtxt);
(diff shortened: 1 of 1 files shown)


main/auth/justification.php
```diff
@@ -177,12 +177,26 @@
...
     $userJustificationList .= $justificationContent.$table->toHtml();
 }
 
+$htmlHeadXtra[] = '<script type="text/javascript" >
+$(function(){
+    $("#justification label").each(function(){
+        var colorG = $(this).css("color");
+        var lgtxt = $(this).text().replace(/ /g,"").length;
+        if (colorG!="green"&&colorG!="rgb(0, 128, 0)"&&lgtxt>3) {
+            alert(lgtxt);
```
NicoDucou, line 194: I think this line should be removed because it generates an alert and open a popup in the navigator with a code in it and it is completely useless and blocking the user and blocking the css final loading of the page when using Chamilo Boost.

LudiscapeXApi (commented):
effectivement j'ai oublié d'enlever le alert(

PR #5028, opened 2023-12-22
Group Category: Add field display_order to entity CGroupCategory

christianbeeznest:


ywarnier:
This (display order) is now managed by resource_node

Issue #5029, opened 2023-12-23
New wiki page failed to be created


mk111:
Ridicolous, a new wiki page simply cannot be created, every attempt ends with
PHP Fatal error:  Uncaught PDOException: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'page_id' cannot be null in /usr/local/www/chamilo/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:117

php 7.4.32, 8.2.12, 8.3.0
mysql 8.0.35
openbsd 7.4, freebsd 14, 

but it's not the case with uniserver, xampp or anything that runs on Windows with any flavou of php !

PR #5030, opened 2023-12-23
fix new wiki page creation

mk111:
fixing
PHP Fatal error:  Uncaught PDOException: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'page_id' cannot be null in /vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:117  #

related to #5029

### DIFF
--- a/main/wiki/wiki.inc.php
+++ b/main/wiki/wiki.inc.php
@@ -374,7 +374,6 @@ class Wiki
             ->setVisibilityDisc(1)
             ->setAddlockDisc(1)
             ->setRatinglockDisc(1)
-            ->setPageId($pageId)
             ->setReflink(trim($values['reflink']))
             ->setTitle(trim($values['title']))
             ->setContent($values['content'])
@@ -388,7 +387,7 @@ class Wiki
             ->setLinksto($linkTo)
             ->setUserIp(api_get_real_ip())
             ->setSessionId($this->session_id)
-            ->setPageId($values['page_id'])
+            ->setPageId(intval($values['page_id']))
             ->setEditlock(0)
             ->setIsEditing(0)
             ->setTimeEdit($time)
@@ -713,6 +712,8 @@ class Wiki
                 $em = Database::getManager();
                 $dtime = api_get_utc_datetime(null, false, true);
 
+                $_clean['page_id'] = $values['page_id'] ?? 0;
+
                 $newWiki = (new CWiki())
                     ->setCId($this->course_id)
                     ->setReflink($_clean['reflink'])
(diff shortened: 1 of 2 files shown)


ywarnier:
Ping @NicoDucou (I think you have worked on this)

main/wiki/wiki.inc.php
```diff
@@ -736,6 +737,7 @@ public function save_new_wiki($values)
                     ->setEditlock(0)
                     ->setIsEditing(0)
                     ->setTag('')
+                    ->setPageId(intval($_clean['page_id']))
```
NicoDucou, line 740: I don't understand why you set the page_id here when the page id will be created once the new wiki is created and then we will assign the page_id base on the id return by this call. This page_id assignment will happen on line 760.

So I would not leave what you put on lines 715 and 749

main/wiki/wiki.inc.php
```diff
@@ -388,7 +387,7 @@ public function save_wiki($values)
             ->setLinksto($linkTo)
             ->setUserIp(api_get_real_ip())
             ->setSessionId($this->session_id)
-            ->setPageId($values['page_id'])
+            ->setPageId(intval($values['page_id']))
```
NicoDucou, line 390: This could be usefull depending on the version of PHP so this could be added.

NicoDucou:
Hi @mk111 
I can not reproduce your problem for me it works fine, but maybe there is a need for a specific configuration to reproduce it.
Anyway checking your PR I did 2 comments. Most important is the first one because I thing your change should not be implemented unless you explain why you think it is needed.
The second that might be the reason for your problem and if you only leave this change we could accept the PR.

Second comment about the error and what is stange is that the column page_id can be null as we can see in src/Chamilo/CourseBundle/Entity/CWiki.php (https://github.com/chamilo/chamilo-lms/blame/62ba004a428bb95dc5c09ab865c7f4539aff41db/src/Chamilo/CourseBundle/Entity/CWiki.php#L50-L55)
```
    /**
     * @var int
     *
     * @ORM\Column(name="page_id", type="integer", nullable=true)
     */
    protected $pageId;
```

And in Mysql we have : 
`  `page_id` int(11) DEFAULT NULL,`

So there should not be an error even if it's null

Best regards

mk111:
Hi @NicoDucou ,
made some futher checking, and, i must agree with your conclusion(s). So, just close the issue.

_(I personally don't like null, it means nothing and does not resolve to nothing. The page id should never be null, even if only for a short period of time between when the record is created and when the record is successively updated)_

PR #5057, opened 2024-01-12
Unit test tweaks

marvil07:
Let us try to iteratively fix phpunit tests.
Also, changes here are a few times skipping assertions, mainly when the actual behavior on the implementation is planned to be changed.

One set of changes are related to adding request data, given some parts of the code assume a request is available, which is naturally not the case on tests.
Once request stack is used all around, that can be mocked, for now, mock what is possible.

### DIFF
--- a/src/CoreBundle/Repository/ResourceRepository.php
+++ b/src/CoreBundle/Repository/ResourceRepository.php
@@ -272,6 +272,8 @@ abstract class ResourceRepository extends ServiceEntityRepository
     {
         $qb = $this->getOrCreateQueryBuilder($qb);
 
+        // TODO Avoid global assumption for a request, and inject
+        // the request stack instead.
         $sessionStudentView = $this->getRequest()->getSession()->get('studentview');
 
         $checker = $this->getAuthorizationChecker();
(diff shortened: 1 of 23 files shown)


ywarnier:
Current state: Tests: 308, Assertions: 1979, Errors: 17, Failures: 8.

ywarnier:
Although we haven't been able to pinpoint the binding issue yet (`Doctrine\ORM\ORMInvalidArgumentException: Binding entities to query parameters only allowed for entities that have an identifier. Class "Chamilo\CoreBundle\Entity\ResourceNode" does not have an identifier.`) (we're lacking time to work on this), there is definitely a link to the display_order field of the resource_node table and the fact that, at some point when deleting a resource, it looks for bound resources and tries to update all the necessary display_order fields (or 'positions' as Doctrine seems to call them internally).

When trying to do that, it gets stuck by the fact that, for some reason, the course doesn't have a resource_node.id, which makes it impossible to call the query updating the display_order, and it fails.

When commenting the `$ea->updatePositions($relocation, $delta, $config);` line in SortableListener.php (because the change of display order due to the course removal happens in the `postRemove` listener, apparently), then the GradeBookCategoryRepositoryTest (the only one I'm testing right now) works fine.

ywarnier:
Incidentally, when I comment the updatePositions() call, I only get 2 errors left:
`Tests: 308, Assertions: 2014, Errors: 2, Failures: 8.`

PR #5070, opened 2024-01-14
User - Fixed Load Error managing User-Exercise Relationships when anonymous is removed - refs BT#21198

christianbeeznest:


christianbeeznest:
I close this PR to have the code with the last version so avoiding conflict

PR #5076, opened 2024-01-14
Exercise - Update Audio Question with Translatable Help Link and Modal - refs BT#21319

christianbeeznest:


ywarnier:
The path is in French, but it is accepted temporarily to get things moving.

PR #5090, opened 2024-01-16
Translation string "LastXMonths" throws "Unknown format specifier"

TheTomcat14:
The translation string for the variable $LastXMonths contains an error, %i must be %d. %i is not a valid format specifier.

The string is used in Statistics::printUsersNotLoggedInStats() in main/inc/lib/statistics.lib.php line 1328


ywarnier:
This was a bit tricky because it should be changed in translate.chamilo.org and not directly here, but I've done that manually so  your PR can be included as is. Sorry for the big lag approving this. Preparing 1.11.28.

PR #5092, opened 2024-01-17
User: Fix usergroup add user behavior for default visibility settings - refs BT#21208

christianbeeznest:


christianbeeznest:
I close this PR to avoid conflicts

PR #5095, opened 2024-01-17
Internal: Migration: Transfer records from track_e_attempt_recording to track_e_attempt_qualify and drop old table

christianbeeznest:


ywarnier:
Following team conversations on the topic, the right use of the track_e_attempt_recording seems to be to record all attempts by students, so track_e_attempt_qualify is not the best renaming on Earth. track_e_attempt_log would be much better.

PR #5096, opened 2024-01-18
User: Refactor anonymous user handling - refs BT#21198

christianbeeznest:


christianbeeznest:
I close this PR to avoid conflicts with last changes in master.

PR #5131, opened 2024-02-01
Internal: User:  Refactor user cascade deletion and handle exceptions - refs BT#21198

christianbeeznest:


christianbeeznest:
I close this PR because is improved in PR 5232

PR #5132, opened 2024-02-01
Update display.lib.php Control if a date is more recent in c_lp_view

LudiscapeXApi:
Control if a date is more recent in c_lp_view (case : skip course home page by direct link to lp in a email)

### DIFF
--- a/main/inc/lib/display.lib.php
+++ b/main/inc/lib/display.lib.php
@@ -1573,6 +1573,30 @@ class Display
             $latestDate = $row['access_date'];
         }
 
+
+        // Control if a date is more recent in c_lp_view (case : skip course home page by direct link to lp in a email)
+        $t_lp_item_view = Database::get_course_table(TABLE_LP_ITEM_VIEW);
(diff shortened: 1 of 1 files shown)


ywarnier:
I didn't review this PR deeply just yet, but for now I'm wondering if seeing the learning path (through a direct link) will not prevent *all* other notifications (from other tools too) to appear...
Because you modify $latestDate directly, and this is the variable then used to scan each tool, one by one.
The idea if you entered the learning path directly is that you *didn't* enter the other tools directly at the same time.
So you should set a $lpLatestDate and then use this one further down in the `foreach ($tools as $tool) { ... $toolName = $tool['name'];` but only when the tool name is `learnpath`. That's the first feeling I get from looking at it.

ywarnier:
Modified and merged in d2772e4b19d99b23bacc438352270a61ca42fb4e

LudiscapeXApi:
Oui pardon effectivement je n'utilise que les learningPath sur mes plateformes Chamilo que je gère, j'ai donc un biais.

PR #5186, opened 2024-02-23
Update trad4all.inc.php

spacecabbie:
The dutch version is missing allot of variables from the main english one. These have now all been added to this file and updates wil follow to fully add the translations.


ywarnier:
Hi @spacecabbie 
I'm sorry I'm seeing this so late (it would have been great if you would have asked or checked our wiki page, although it was not super clear either, before doing such a large change). (https://github.com/chamilo/chamilo-lms/wiki/Coding-conventions#translations-management)

The terms have to be translated in our translation system at https://translate.chamilo.org, for which you need to create an account and ask for approval (you can ask me that).

The format of your PR makes it very work-intensive (some variables are translated to English rather than Dutch, which is unnecessary given the fact Chamilo always falls back to English if the translation is not available).

spacecabbie:
> Hi @spacecabbie I'm sorry I'm seeing this so late (it would have been great if you would have asked or checked our wiki page, although it was not super clear either, before doing such a large change). (https://github.com/chamilo/chamilo-lms/wiki/Coding-conventions#translations-management)
> 
> The terms have to be translated in our translation system at https://translate.chamilo.org, for which you need to create an account and ask for approval (you can ask me that).
> 
> The format of your PR makes it very work-intensive (some variables are translated to English rather than Dutch, which is unnecessary given the fact Chamilo always falls back to English if the translation is not available).

I am Sorry Yannick,
I knew the translate system is prefered, But I did not know that making a PR gives more work i do apoligize for that I will resubmit via the translate system.

spacecabbie:
> Hi @spacecabbie I'm sorry I'm seeing this so late (it would have been great if you would have asked or checked our wiki page, although it was not super clear either, before doing such a large change). (https://github.com/chamilo/chamilo-lms/wiki/Coding-conventions#translations-management)
> 
> The terms have to be translated in our translation system at https://translate.chamilo.org, for which you need to create an account and ask for approval (you can ask me that).
> 
> The format of your PR makes it very work-intensive (some variables are translated to English rather than Dutch, which is unnecessary given the fact Chamilo always falls back to English if the translation is not available).

I am Sorry Yannick,
I knew the translate system is prefered, But I did not know that making a PR gives more work i do apoligize for that I will resubmit via the translate system.

The reason i did it like this is because the translation system does not export untranslated variables so its hard to do a batch job on translations or see what is missing. With google translate being so good these days i basicaly pull the full file trough that and then proof read all the translations saving loads of time.

It would be really helpfull if all translation files regardless of lang have untranslated variables in english so the files stay uniform in amount of lines and variable positions. I could easly translate German as wel then ontop of dutch.

PR #5187, opened 2024-02-25
Update trad4all.inc.php

spacecabbie:
Deleted dublicate variables


ywarnier:
Not really sure what this PR used to contain (it is not visible clearly on Github), but changes to translation files have to go through translate.chamilo.org.

Issue #5221, opened 2024-03-07
C2: Create "fallback" user role and account to attach orphan content to


ywarnier:
When (really) deleting a user in C2, all related resources are deleted at the moment.
This is not the desired behaviour, but due to foreign keys, we don't really have a choice to leave a large number of objects with a null reference. This would make them really difficult to clean-up afterwards.
Instead, we are going to use the "hack" of creating a FALLBACK role and a 'fallback' user that cannot be deleted.

Whenever we delete user 123, we will scan the contents linked to that user, and replace the reference to the id of the FALLBACK user.
This way, all content of user 123 is not deleted, only reassigned.

This role and user should be created on install and on migration from 1.11.

christianbeeznest:
While implementing the solution to reassign resources from deleted users to a fallback user, I've realized I might encounter unique constraint violations in the database. This is particularly true for tables with unique constraints involving the user ID and other fields.

For example : 

In SessionRelUser 

#[ORM\UniqueConstraint(name: 'session_user_unique', columns: ['session_id', 'user_id', 'relation_type'])]

UserRelUser

#[ORM\UniqueConstraint(name: 'user_friend_relation', columns: ['user_id', 'friend_user_id', 'relation_type'])]

Does anyone have suggestions on how to anticipate and handle these conflicts effectively? I'm considering reviewing the affected entities and relationships and adding specific tests to ensure database integrity. Any advice on balancing data preservation with database performance and integrity would be greatly appreciated.

ywarnier:
Regarding the listing of affected entities and relationships, I would suggest just scanning the entities for any user_id reference.

Regarding anticipating these conflicts : automated testing is the answer, although these are complex tests, there could be just one big test to test all of this: go through Chamilo, create some stuff as a new user, then delete the user and go around the tools as admin to see if an error pops up...

This being said, I don't think you would hav a lot of constraint integrity issues if you just *update* an object to point to a new user_id. The constraint will protest if you *delete* a record (which is what we are trying to avoid), not if you update it.

Any additional opinion, @AngelFQC ?

ywarnier:
OK, so to be clear and thanks to @christianbeeznest 's list of user fields in the database, here is a list of all fields we should keep (replace with fallback) or delete on user deletion.

The general rule is that a field should be *converted* to use the fallback user ID if it relates to some *content* we don't necessarily want to delete, or important history that might affect other users (like text feedback or grading on some other user's work). In any other case, we should delete the record.

[...]

PR #5232, opened 2024-03-11
User: Create "fallback" user role and account to attach orphan content to - refs #5221

christianbeeznest:


src/CoreBundle/Entity/TicketCategoryRelUser.php
```diff
@@ -26,5 +29,5 @@ class TicketCategoryRelUser
 
     #[ORM\ManyToOne(targetEntity: User::class)]
     #[ORM\JoinColumn(name: 'user_id', referencedColumnName: 'id', onDelete: 'CASCADE')]
-    protected TicketCategory $user;
+    protected User $user;
```
ywarnier, line 32: :+1:

ywarnier:
Code looks good. I was expecting more changes to entities' definition, but maybe it makes sense.

ywarnier:
When I execute `php bin/console doctrine:schema:update --force --complete`, the migration is not executed (or at least the user is not created). When I try `php bin/console doctrine:migrations:execute Version20240310160200`, it tells me `Migration class "Version20240310160200" was not found?`
Not sure how else to do this.

PR #5234, opened 2024-03-11
Course: Fix csv import: Preserve existing non-required fields - refs BT#21441

christianbeeznest:


main/admin/course_update_import.php
```diff
@@ -102,17 +102,39 @@ function validateCourseData(array $courses): array
 function updateCourse(array $courseData, int $courseId): void
 {
     $courseTable = Database::get_main_table(TABLE_MAIN_COURSE);
-    $params = [
-        'title' => $courseData['Title'],
-        'course_language' => $courseData['Language'],
-        'category_code' => $courseData['CourseCategory'],
-        'visual_code' => $courseData['Code'],
+    $params = ['visual_code' => $courseData['Code']];
```
ywarnier, line 105: The visual_code should not be changed through this CSV import, you can just remove that line.

ywarnier (changes requested):

main/admin/course_update_import.php
christianbeeznest: ok, it is removed that line now.

ywarnier (approved):

PR #5237, opened 2024-03-12
Reports - Improvements on message tracking

juancp-contidosdixitais:
Currently, tracking messages from the user report would only indicate cases when the user receives a message, not when the user sends a message, since it is assumed by default that a sent message will be answered, but this is not always the case. If we want to fully track, we must include those sent messages.

image

We created a new method (getMessageExchangeWithUser) in main/inc/lib/message.lib.php to obtain both the users from whom they have received messages and to whom they have sent messages and we modified the loading of this data in main/mySpace/myStudents.php.

On the other hand, a bug was identified where the message load is duplicated:

https://github.com/chamilo/chamilo-lms/blob/5041118ba52005fa85c47da68335c1503681429c/main/mySpace/myStudents.php#L2404-L2409

### DIFF
--- a/main/inc/lib/message.lib.php
+++ b/main/inc/lib/message.lib.php
@@ -3073,7 +3073,7 @@ class MessageManager
                  WHERE user_receiver_id = $userId" .
                  ($startDate ? " AND send_date >= '$startDate'" : "") .
                  ($endDate ? " AND send_date <= '$endDate'" : "") .
-               "UNION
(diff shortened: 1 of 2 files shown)


ywarnier:
Looks good to me. I was worried it might generate a lot of load on the system, but this is *inside* a check on `api_get_configuration_value('allow_user_message_tracking')`, so not many people will need it.

Issue #3795, opened 2021-02-23
WYSIWYG Editor filemanager (tinymce)


jmontoyaa:
Implement the filemanager in ckeditor using resources

jmontoyaa:
This was replaced with tinymce and it works when open the editor inside a course.

Filemanager doesn´t work yet outside a course.

jmontoyaa:
https://github.com/chamilo/chamilo-lms/commit/5f6a97ca5ae6ef34a787b1c201b78a3797387fc9

jmontoyaa:
The personal files (#3792)  should be connect to the filemanager.

christianbeeznest:
It's added an image picker for tinymce in courses tools (no documents) with code legacy in this commit https://github.com/chamilo/chamilo-lms/commit/ef3c8b0b15684ead7298bdb937ec910c19599994 

For tinimce with vuejs inside course is added this commit https://github.com/chamilo/chamilo-lms/commit/5f6a97ca5ae6ef34a787b1c201b78a3797387fc9  , It should be add in other tools when it is using vuejs,

For tiniymce outside a course with vuejs to use the filemanager created for tool documents (vuejs) changing the browser url by personal files [/resources/personal_files/]

christianbeeznest:
It is added editor filemanager for documents and messages in this PR https://github.com/chamilo/chamilo-lms/pull/4756

NicoDucou:
The tinyMCE works on many page.
The missing point are extension that we had before like mathjack but probably it could be removed for 2.0 and reintegrated later.

ywarnier:
Work on this with @daniboygg to generalize the implementation of the filemanager in TinyMCE:
- as a component on top of Vue components
- that supports the same/similar options as toolbars (as were previously used by CkEditor - see CkEditor.php)
- that is configurable per tool

ywarnier:
Reference code: https://github.com/chamilo/chamilo-lms/commit/c111d5d160a326d4ce2e69c1065aae2fb2c1f5d1

ywarnier:
At the moment this saves the image as base64 in the database. The images should be saved on disk in var/[tool-name]/[uuid].[extension].

ywarnier:
- wrapper for tinymce
- configuration of plugins
- upload images directly (not base64)

daniboygg:
I've done a little research about uploading files. 

[...]

PR #5273, opened 2024-03-21
Gh 3795 editor filemanager

daniboygg:
Use BaseTinyEditor in all places in Vue code to standardize the code across all project. This fixes https://github.com/chamilo/chamilo-lms/issues/3795


AngelFQC:
@daniboygg  I've sent two additional commits (e8fd9fcdd4502a672063b4fa972d1e87a61f71d2 and 981eabaaef97fe5f931aa37c01f34b1b7f0defdb) to allow select files according to their file types and fix the call ajax

PR #5320, opened 2024-04-01
[C2] Tinymce problems in development

daniboygg:
Modify documentation about how to run development Vue server in a corner case when dev-server cannot be used. 
Enable fullpage plugin of tinymce in toolbar options, otherwise this plugin will not be shown to the user.

@AngelFQC using the command dev-server the plugin TiniMCE still make the request to the server with the default url instead of using http://localhost:8000. 

For now, I documented this case because I think this will need to be addressed in encore, and I'm not sure is solvable.

Do you have any idea if encore could change the request made by the plugin? (The requests that are failing in development are the ones made to load the skins CSS of TinyMCE, they are loaded from the script Editor.js that I think lives in the plugin)

These are the lines where we tell encore to load these files from server, and using dev-server the new url serving the assets is not taken into account
```
Encore.copyFiles({
  from: "./node_modules/tinymce/skins",
  to: "libs/tinymce/skins/[path][name].[ext]",
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,10 +155,14 @@ Load the (your-domain)/main/install/index.php URL to start the installer (which 
 If the installer is pure-HTML and doesn't appear with a clean layout, that's because you didn't follow these instructions carefully.
 Go back to the beginning of this section and try again.
 
-If you want hot reloading for assets use the command `yarn run encore dev-server`. This will refresh automatically
-your assets when you modify them under `assets/vue`. Access your chamilo instance as usual. In the background, this will serve
+If you want hot reloading for assets, use the command `yarn dev-server`\*. This will refresh automatically
+your assets when you modify them under `assets/vue`. Access your Chamilo instance as usual. In the background, this will serve
 assets from a custom server on http://localhost:8080. Do not access this url directly since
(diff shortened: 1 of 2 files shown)


AngelFQC:
Hi @daniboygg 

For now, I have disabled the skin and content_url settings so that tinymce can work with dev-server
The skin will be included within the compiled css of the app (94108b963e6268d7426f84ac9ffde7d9302634cb)

I have commented out the lines that should work to include the css content, but I can't get it to work with dev-server
But it is needed to include content_url, because in 1.11.x there is user theme functionality that can include your own css in the editor

And I have also cherry-picked your commit related to full-page (d162a79a7b7c7b7036bf83938334c202f66bb297)

daniboygg:
> I have commented out the lines that should work to include the css content, but I can't get it to work with dev-server
But it is needed to include content_url, because in 1.11.x there is user theme functionality that can include your own css in the editor

The error only happens with `encore dev-server` command, with `encore dev` it seems to work. Could be possible to disable skin and content_url settings only for this command, does encore offer some environment variable to know if we are in dev-server, dev or production mode?

Issue #5313, opened 2024-03-31
C2: Settings current titles, comments and migrations


ywarnier:
In the current dev version of Chamilo 2 (i.e. master branch), there are issues with the "title" and "comment" columns of the settings_current table.

In Chamilo 1.*, the `institution` variable defined title `InstitutionTitle` and comment `InstitutionComment` as names for language variables.
However, these variables have now been converted in Chamilo 2 to the *values* of the translation in English.
So `InstitutionTitle` has become `Organization name`, and `InstitutionComment` has become `The name of the organization (appears in the header on the right)`.

The old translation files have now been removed from C2's repository (due to he fact that we didn't catch the missing comments) and thus the relationship between the `InstitutionTitle` variable name and the `Organization name` gettext ID is now lost unless we re-use them to create the missing variables. We would only need the old main/lang/english/trad4all.inc.php to get the link back, though, and only to *prepare* the migration/install (so it doesn't need to be permanently added back to Git).

So now we have 2 "situations" to clean up:
1. the settings_current table should change the definition of the title and comment fields to "text" (instead of varchar(255))
2. fresh installs of Chamilo 2 create settings_current where title = variable. This is *not* correct. 
2.1. When the variable already existed in Chamilo 1, it should use the English translation of the language variable for that settings_current.title in the title column, and the English translation of the language variable for that settings_current.comment in the comment column. For example, variable "administrator_email" should use the English translation of the old "title" column (emailAdministratorTitle), which is "Portal Administrator: e-mail" (this variable already exists in Gettext, so just use `get_lang()` to translate it on screen), and the "comment" was "emailAdministratorComment" and should now be "The e-mail address of the Platform Administrator (appears in the footer on the left)" (already exists in Gettext).
2.2. When the variable did not exist in Chamilo 1 (in settings_current) but it was already present in configuration.dist.php, then it is more complicated (you can use the list I will create below for that).
2.3. If the variable was added in C2 (did not exist anywhere previously), then please report it so I can provide a title and comment for them.
3. migrations from Chamilo 1 maintain the "title" and "comment" fields as they were in Chamilo 1, so these have to be converted to gettext format. An update table should be setup in the migration to convert "InstitutionTitle" to "Organization name" and "InstitutionComment" to "The name of the organization (appears in the header on the right)". This can probably be done by scanning main/lang/english/trad4all.inc.php real quick and building a table then copy-pasting only the ones used in settings_current.title or .comment it in the code of the migration.

[...]

PR #5345, opened 2024-04-09
Internal: Migration: Update title and comment values in settings_current - refs #5313

christianbeeznest:


christianbeeznest:
I close this PR because is improved in other one.

PR #5350, opened 2024-04-10
Gh 5341 save colors with name

### LINKED ISSUE #5341: [C2] Save colors with a name
Save colors defined in admin with a name in backend.

Show preview of components on the right side, and make the color pickers smaller

daniboygg:
Fix #5341 

Update color selection page in Chamilo

image

Right now, the endpoints of the backend allow storing a theme with a name. I retrieve the list of themes and show them in a selector. Change the theme a press on save, store the theme in the backend as the active theme.

The backend does not allow overriding a theme, creates a new every time.

To allow the user to define their custom theme, there are a couple of options:
1. Make a button like "Create new theme with these colors" that shows a text field to set a name and a save button
2. Edit the dropdown in place. Vue prime allows to make a field editable https://primevue.org/dropdown/#editable. The flow would be:
    * Is user select a theme, the theme changes, if the user click on save theme will be set as active.
    * If the user writes something in the field, a click on save would save the theme with the name written.

I'm not sure about the 2 one, will the user notice that the field is editable?
  
Right now there is no default theme, so the theme selector shows no option. There will be some themes in database Chamilo automatically? Or do I need to change something in the frontend to show the first theme?

### DIFF
--- a/assets/vue/components/basecomponents/BaseColorPicker.vue
+++ b/assets/vue/components/basecomponents/BaseColorPicker.vue
@@ -5,11 +5,12 @@
       <ColorPicker
         format="hex"
         :model-value="hexColor"
+        class="w-11"
         @update:model-value="colorPicked"
       />
       <BaseInputText
         label=""
-        class="max-w-32"
+        class="w-32"
         input-class="mb-0"
         :model-value="hexColor"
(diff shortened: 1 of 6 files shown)


daniboygg:
@ywarnier @AngelFQC

PR #5384, opened 2024-04-17
Remove vuex from simple components

daniboygg:
Remove vuex from components where is imported but not used

#5377

### DIFF
--- a/assets/vue/components/assignments/AssignmentsLayout.vue
+++ b/assets/vue/components/assignments/AssignmentsLayout.vue
@@ -5,14 +5,12 @@
 <script setup>
 import { computed, ref, watchEffect } from "vue";
 import { useRoute, useRouter } from "vue-router"
-import { useStore } from "vuex";
 import { useI18n } from "vue-i18n";
 import { useCidReq } from "../../composables/cidReq";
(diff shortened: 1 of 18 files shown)


ywarnier:
@daniboygg there's a small conflict now and I'm not clear whether these 2 lines should remain or not.
One (`const { isAdmin } = useSecurityStore()`) was added by you in a more recent PR, so I suppose it has to stay, and the other (`const store = useStore()`) was added by Christian 2 months ago to recover some features in the social network.
I'm leaving these 2 lines there, but don't hesitate to send another PR if this is introducing something you wanted to remove.

daniboygg:
For what I can see this line `const store = useStore()` will fail now, because there is no import in the component that imports the function useStore here https://github.com/daniboygg/chamilo-lms/blob/046dda11a575866cbb499fd34a49e4a4433d7e72/assets/vue/components/social/UserProfileCard.vue#L112. Should I make a PR as soon as possible, addressing this? I think this change will break every page where this component is used otherwise

In this line `const { isAdmin } = useSecurityStore()` I used the new useSecurityStore from pinia, and it's the correct one from now on.

PR #5385, opened 2024-04-17
[C2] Delete some components that are not used

daniboygg:
This will remove components that are not in use in Vue application. This will delete some unneeded vuex dependencies.

#5377


ywarnier:
I'm not able to review this one, in particular because of the assets/vue/router/course.js removal. I think *maybe* the corresponding views are a work in progress to be used later. @AngelFQC if these are not works in progress, can we remove them and implement them correctly later on, or should we keep them there to hold the place for the file to be used later?

assets/vue/router/course.js
```diff
@@ -1,29 +0,0 @@
...
-    {
-      name: 'CourseList',
-      path: '',
-      component: () => import('../views/course/List.vue')
-    },
-    {
-      name: 'CourseCreate',
-      path: 'new',
-      component: () => import('../views/course/Create.vue')
-    },
```
AngelFQC, line 17: CourseCreate is currently in use

assets/vue/views/course/Create.vue
AngelFQC, line 1: This page is used to allow teachers to create their courses  (from the orange button on the /courses page)

AngelFQC (changes requested):

assets/vue/views/course/Create.vue
```diff
@@ -31,18 +30,17 @@ import courseService from "../../services/courseService"
...
-const store = useStore()
 const item = ref({})
 const router = useRouter()
 const { t } = useI18n()
 
-const isLoading = computed(() => store.getters['course/getField']('isLoading'))
-const violations = computed(() => store.getters['course/getField']('violations'))
-const courseData = ref({})
+const isLoading = ref(false)
+const violations = ref(null)
```
daniboygg, line 38: The reference violations is used in the component CourseForm but this component has no property errors...
```html
<CourseForm
      ref="createForm"
      :errors="violations"
      :values="item"
      @submit="submitCourse"
    />
```

daniboygg:
I reverted the changes. So, do we need the components CourseList, CourseUpdate and CourseShow. All of them are using vuex, but when I visit the route I see nothing. 
I think we should fix these components before remove vuex, otherwise it will be difficult to ensure the behavior is consistent with the old one.

ywarnier:
Tested. I don't see anything missing, feature-wise, by testing it with this PR. Will include soon in master.

PR #5396, opened 2024-04-18
Plugin: Keycloak: Update README

ivosh:
Fix README so that the mapper for the first name is what is expected by the code:
https://github.com/chamilo/chamilo-lms/blob/1.11.x/plugin/keycloak/start.php#L108


ywarnier:
Merged. Thank you !

PR #5436, opened 2024-04-25
Fallback language hack

ywarnier:


ywarnier:
Before change:
before-lang-hack
After change:
after-lang-hack

PR #5485, opened 2024-05-06
Internal: Make a request to reset the current course

AngelFQC:


ywarnier:
This is left aside for now as we don't measure its complete impact yet.
Requires more analysis and testing.

PR #5540, opened 2024-05-23
Server session: New expiration notification system

juancp-contidosdixitais:
Some teachers use lessons with a minimum duration time that requires students to stay connected for a certain period to progress and move on to the next lesson.

The problem arises when a student leaves the lesson inactive and returns to it after some time, usually a long time. When they do, their progress timer still displays the elapsed time, but in reality, if the session lasts, for example, two hours and eight hours have passed, the student is unaware because they do not receive a notification that their session has expired.

We have built a session expiration system that, after a certain amount of time, displays a screen indicating the session has expired, and even shows a countdown in the last few seconds.

Initially, this would require a subprocess that checks every X seconds for the remaining session time. This subprocess would need to be analyzed to determine if its execution would extend the session, which would be an undesirable effect.

When a user's session has expired, the system hides the entire campus with a black screen and indicates the end of the session.

This PR references this issue: #4776 

image

This session control can be enabled from configuration.php by setting the 'session_lifetime_controller' option to true.

### DIFF
--- a/main/inc/ajax/session_clock.ajax.php
+++ b/main/inc/ajax/session_clock.ajax.php
@@ -1,0 +1,75 @@
+<?php
+
+require_once __DIR__.'/../../../vendor/autoload.php';
+require_once __DIR__.'/../../../app/AppKernel.php';
+
+$kernel = new AppKernel('', '');
+
+// Check for 'action' parameter in the GET request
+if (isset($_GET['action'])) {
+    $action = $_GET['action'];
+
+    if ($action == 'time') {
+        // Load the Chamilo configuration
+        $alreadyInstalled = false;
+        if (file_exists($kernel->getConfigurationFile())) {
+            require_once $kernel->getConfigurationFile();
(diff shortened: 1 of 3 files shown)


ywarnier:
Please always indicate the language variables you are adding ;-)

ywarnier:
Don't use "JS" as a suffix. If you need to escape a variable to use it for JavaScript code, simply use the `| escape('js')` in the Twig template, which you have already done.

ywarnier:
When I enable this PR with a 60s session_lifetime in configuration.php, I do get the SessionExpiredAtJS message on a black background, but when I click the button I am immediately connected again with the counter getting back to 55s.

I think it doesn't like the 60s timeout because it's also the limit at which it should show the warning.

ywarnier:
Merged here: https://github.com/chamilo/chamilo-lms/commit/14e699f23ceaec8dd64e21ccf1fc8bc09c58010d
Thanks!

PR #5541, opened 2024-05-23
User registration: Enable extra field validation for user registration in Chamilo

juancp-contidosdixitais:
This pull request introduces a new functionality to Chamilo that allows the use of a user extra field to validate user registrations. For instance, if we have an extra field to store passport numbers, we can ensure that there are no duplicate values.

registro_manual

This functionality also works on batch user registrations.

resultado_importacion

To utilize this functionality, you need to specify in "extra_fields_to_validate_on_user_registration" within configuration.php the extra fields you want to validate. For example:

$_configuration['extra_fields_to_validate_on_user_registration'] = [
    'extra_fields' => [
        'passport', 'employee_id'
    ]
];

### DIFF
--- a/main/admin/user_add.php
+++ b/main/admin/user_add.php
@@ -368,137 +368,176 @@ $html_results_enabled[] = $form->createElement('button', 'submit_plus', get_lang
 $form->addGroup($html_results_enabled);
 
 // Validate form
-if ($form->validate()) {
+$formValid = $form->validate();
+if ($formValid) {
     $check = Security::check_token('post');
     if ($check) {
         $user = $form->exportValues();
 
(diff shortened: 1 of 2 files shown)


ywarnier:
I see the point of this PR, but the code comparison is a complete mess (probably not your fault - some mix of bad detections by Github and of adding changes to many small places, and maybe some parallel history of the same files in the last 5 months after your PR), so in the current state I cannot safely approve it (it would have to be rewritten with less intrusive changes).

ywarnier:
This has been integrated (slightly differently) in https://github.com/chamilo/chamilo-lms/commit/75ef74554322f26ed883433e0460e212f66210af in mid-2025

PR #5543, opened 2024-05-24
Session configuration: Option to hide visibility options after end date

juancp-contidosdixitais:
This pull request adds an option in configuration.php to configure visibility options after the end date in a session.

image

It is configured by indicating which options we want to hide, which can be three, SESSION_VISIBLE_READ_ONLY, SESSION_VISIBLE or SESSION_INVISIBLE

image

The end result is that the configuration combo will not show the indicated options

image


ywarnier:
Merged in https://github.com/chamilo/chamilo-lms/commit/c8748f86f8a409cbd1e1a7195bbcf8ff81a28cae
Thanks!

PR #5557, opened 2024-05-29
Exercise list actions: Restrict Actions for Hidden Exercises in Base Course

juancp-contidosdixitais:
When an exercise is deactivated in the base course (indicated by a closed eye icon), Chamilo does not allow its visibility to be changed from a session. Despite this, the system allows clicking on the icon, and although it does not modify the state of the exercise, it gives the impression that it does, even displaying a success message after the action.

image

This pull request ensures that if an exercise is hidden in the base course, its icon is not clickable. Additionally, for bulk actions, it checks if there are hidden exercises in the base course and, if so, indicates that the status of these cannot be changed.

image

### DIFF
--- a/main/exercise/exercise.class.php
+++ b/main/exercise/exercise.class.php
@@ -9909,15 +9909,29 @@ class Exercise
                                 );
                             } else {
                                 if ($row['active'] == 0 || $visibility == 0) {
-                                    $visibility = Display::url(
-                                        Display::return_icon(
-                                            'invisible.png',
-                                            get_lang('Activate'),
-                                            '',
-                                            ICON_SIZE_SMALL
-                                        ),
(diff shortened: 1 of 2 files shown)


ywarnier:
The message in the screenshot does not match your code (the message does not "include" the exercise title as in your screenshot).
Also... it does not seem to work to show the message because your change alters the "visible" action. In the exercises list, the "eye" icon sends the "choice=disable" param instead of setting the "action=visible" param.

This PR seems incomplete in that respect. I have added the `CannotChangeVisibilityOfBaseCourseResourceX` to the languafe files. Should be used as `Display::return_message(sprintf(get_lang('CannotChangeVisibilityOfBaseCourseResourceX'), $objExerciseTmp->name)`

ywarnier:
@juancp-contidosdixitais ping

ywarnier:
The PR has been updated and merged into 751175aee528ef4e0bcb284697eb359998ad1341

PR #5565, opened 2024-06-03
Add migration script to delete unnecessary shared_folders for OFAJ - refs BT#21680

christianbeeznest:


ywarnier:
Looks good to me, but difficult to test.

NicoDucou:
Applied for the specific case and not necessary in Chamilo Master because it has been implemented in the migration process directly.

PR #5593, opened 2024-06-17
Internal: Improve partial tinyMCE filemanager wiht legacy - refs BT#21647

christianbeeznest:


christianbeeznest:
I close this PR due conflicts , it will be create other version.

PR #5614, opened 2024-06-25
Changes about platform style theme - refs BT#21621

AngelFQC:


public/main/template/default/gradebook/custom_certificate.html.twig
```diff
@@ -9,30 +9,30 @@
 <table border="0" cellpadding="0" cellspacing="0">
     <tr>
         <td colspan="3">
-            <img src="{{ url('index') ~ 'build/css/themes/' ~ stylesheet_name ~ '/images/cuadro.png' }}">
+            <img src="{{ theme_asset('images/cuadre.png') }}">
```
ywarnier, line 12: Not sure where cuadre.png or cuadro.png come from. I cannot see them as files in a default installation...

NicoDucou: These are the files used for the general certificate from OFAJ (which is custom as indicated in the name of the template).

PR #5629, opened 2024-07-03
LRS fixes from 1.11.x

AngelFQC:
See #3943


ywarnier:
Not entirely reviewed but merging as we'll get time to review this in the following months and we need the database changes to close the alpha.

Issue #5644, opened 2024-07-08
[C2] Roles and permissions fine-grained management


ywarnier:
Add the possibility to define fine-grained permissions and assign them to roles.

A definition of basic permissions is ongoing here: https://github.com/chamilo/chamilo-lms/wiki/Permissions

Create a Permission entity with id, title, slug and description.
Create a PermissionRelRole entity with id, permission_id, role_code, changeable, updated_at.
Create an assignment interface with roles in columns and permissions in rows, like so:

| Permission | ROLE_INVITEE | ROLE_STUDENT | ROLE_TEACHER ... |
| ---- | ---- | ---- | ---- |
| analytics:view |  [] | [x] | [x] |
| assignment:create	| [] | [] | [X] |

allowing one to check the boxes for each role (except when changeable == 0). This is then sent to the PermissionRelRole entity/table.

Use the wiki table as source for the default definition of each set of permissions per role.

Define a global permission function to search for the existence of a permission based on the permission name and an array of roles. Something like `api_get_permission('analytics:view', ['ROLE_STUDENT','ROLE_TEACHER'])`. This function/helper will then check, for the given permission, if *one or more* of the roles provided has the permission enabled and return true or false based on that.

Relates to #2540

christianbeeznest:
Added in this PR https://github.com/chamilo/chamilo-lms/pull/5651  , waiting for testing to merge , the interface for permissions management is /permissions

ywarnier:
Well done!

ywarnier:
@christianbeeznest 
We need to add one permission "Edit course settings" so that teachers can be given limited access to a course, if needed. This permission is *on* by default for all teachers, but can be removed (also for all teachers) at once.

christianbeeznest:
It is done with this commit https://github.com/chamilo/chamilo-lms/pull/5723/commits/70b8eef2bd90d3ad518da5a2367c994d615b9f39

ywarnier:
There's still a lot to do to make this funcional for final users, but the goal of the issue has been achieved, although some comments about optimizing remain unattended in the PR like https://github.com/chamilo/chamilo-lms/pull/5651/files#r1672376049 and https://github.com/chamilo/chamilo-lms/pull/5651/files#r1671756095. These can be attended later on, before the stable.

PR #5651, opened 2024-07-10
Internal: Add permission management and role assignment functionality - refs #5644

christianbeeznest:


src/CoreBundle/Controller/PermissionController.php
```diff
@@ -0,0 +1,100 @@
...
+    #[IsGranted('ROLE_ADMIN')]
+    #[Route('/permissions', name: 'permissions')]
+    public function index(
+        PermissionRepository $permissionRepo,
+        PermissionRelRoleRepository $permissionRelRoleRepo,
+        Request $request,
+        EntityManagerInterface $em
+    ): Response {
+        $permissions = $permissionRepo->findAll();
+        $roles = ['ROLE_INVITEE', 'ROLE_STUDENT', 'ROLE_TEACHER', 'ROLE_ADMIN', 'ROLE_SUPER_ADMIN', 'ROLE_GLOBAL_ADMIN', 'ROLE_RRHH', 'ROLE_QUESTION_MANAGER', 'ROLE_SESSION_MANAGER', 'ROLE_STUDENT_BOSS'];
```
ywarnier, line 48: We should have a method somewhere that returns all supported roles. Having this list hardcoded here is probably not a good idea.

src/CoreBundle/Migrations/Schema/V200/Version20240709222600.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Create permissions and permission_rel_roles tables';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('
+            CREATE TABLE IF NOT EXISTS permissions (
```
ywarnier, line 22: Tables in Chamilo have to be named in singular (`permission`, not `permissions`). This is an old standard that we try to maintain in all contexts except when it would bring confusion vs something else.

src/CoreBundle/Migrations/Schema/V200/Version20240709222600.php
```diff
@@ -0,0 +1,51 @@
...
+                title VARCHAR(255) NOT NULL,
+                slug VARCHAR(255) NOT NULL,
+                description LONGTEXT DEFAULT NULL,
+                UNIQUE INDEX UNIQ_2DEDCC6F989D9B62 (slug),
+                PRIMARY KEY(id)
+            ) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB ROW_FORMAT = DYNAMIC
+        ');
+
+        $this->addSql('
+            CREATE TABLE IF NOT EXISTS permission_rel_roles (
```
ywarnier, line 33: Same with relations: `permission_rel_role`, not `permission_rel_roles`

src/CoreBundle/Service/PermissionService.php
AngelFQC, line 1: Mejor ponerlo en ServiceHelper 😅

PR #5657, opened 2024-07-11
Elfinder  file manager: Fix Multi-URL Mode Document Access Issue in Chamilo 1.11.x

juancp-contidosdixitais:
Chamilo 1.11.x in multi-url mode has a minor flaw that may allow access to documents from other sessions through the elfinder file manager.

This issue could cause users from sessions from different children sites sharing the same base course view all documents uploaded regardless of the session because elfinder cannot distinguish which session a document belongs to.

**Steps to Replicate:**

1. Create a base course and upload Document A.
2. Create Session 1, add the course to the session, and upload Document B.
3. Create Session 2, add the course to the session, and upload Document C.

Now, from the elfinder file manager, (any instance of the CKEditor for example), we should only see Documents A and B. However, Documents A, B, and C are visible.

**Proposed Mitigation:**
This PR implements the following actions to mitigate the issue:

1. When files are loaded in the file manager, it checks if the files belong to the current session (those that have a session number pattern like "{session_number}" in the name). If they do not belong to the session, the file manager icons are removed.
2. At the time of attempting to manage any of the files, it verifies if the file belongs to the current session checking the name pattern,. If it does not belong, interactions with the file are not allowed.

### DIFF
--- a/main/inc/lib/elfinder/filemanager.php
+++ b/main/inc/lib/elfinder/filemanager.php
@@ -18,8 +18,10 @@ if (file_exists($file)) {
     $language = $iso;
 }
 $questionId = isset($_REQUEST['question_id']) ? (int) $_REQUEST['question_id'] : 0;
+$sessionId = api_get_session_id();
 
 $template->assign('question_id', $questionId);
+$template->assign('session_id', $sessionId);
 $template->assign('elfinder_lang', $language);
 $template->assign('elfinder_translation_file', $includeFile);
 $template->display('default/javascript/editor/ckeditor/elfinder.tpl');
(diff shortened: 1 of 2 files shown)


ywarnier:
I believe this is incorrect, as the files may have other context items surrounded by `__`.
For example, if you use the "Groups" tool inside a course and someone uploads a group document there, you will have something like `Comparison__10__11.csv` where `10` is the session ID and `11` is the group ID. So your comparison pattern has to be a little more complex...

ywarnier:
The code had to be adapted a little (as per my previous comments) but it has been integrated, ready for 1.11.34

PR #5706, opened 2024-08-04
Internal: Fix user creation and update logic for CSV sync script - refs BT#21895

christianbeeznest:


christianbeeznest:
I close this PR to avoid conflict with changes in db from master

PR #5783, opened 2024-09-06
Course: Implement MBZ file generation for Moodle 3/4 backups (single-choice quizzes)

christianbeeznest:


main/coursecopy/export_moodle.php
```diff
@@ -0,0 +1,141 @@
...
+    } else {
+        echo Display::return_message(get_lang('NoResourcesToExport'), 'warning');
+    }
+} else {
+    $form = new FormValidator(
+        'create_export_form',
+        'post',
+        api_get_self().'?'.api_get_cidreq()
+    );
+    $form->addElement('header', get_lang('SelectOptionForExport'));
```
ywarnier, line 102: This block header is not necessary. The setting can go directly under the selection of export type.

ywarnier (changes requested):

main/coursecopy/export_moodle.php
```diff
@@ -0,0 +1,170 @@
...
+        $adminEmail = $_POST['admin_email'];
+
+        $exporter = new MoodleExport($course);
+        $exporter->setAdminUserData($adminId, $adminUsername, $adminEmail);
+
+        // Perform export
+        $courseId = api_get_course_id();
+        $exportDir = 'moodle_export_' . $courseId;
+        try {
+            $moodleVersion = isset($_POST['moodle_version']) ? $_POST['moodle_version'] : '3';
```
ywarnier, line 68: If we known $moodleVersion must be an integer value, we must ensure $_POST['moodle_version'] is an int (cast as (int), for example)

ywarnier (changes requested):
A few things have to be improved. Filtering and documentation, mostly.

Issue #4431, opened 2022-10-23
Add TOTP 2FA authentication support


ywarnier:
Chamilo 1.11.16 does not support 2FA, which is increasingly asked for by large organizations.

Enable 2FA using TOTP (probably the easiest and most commonly used protocol for that) using https://github.com/Spomky-Labs/otphp/blob/v10.0/doc/index.md (v10 is the last one supporting PHP < 8.1).

The implementation probably requires a private key, which should be stored in an extra field (totpprivkey), in a reversible encrypted form, using the user's own salt.

Allow the feature to be "optional" or "enforced". If optional, users can simply enable the feature in a separate page on their profile page. If enforced, users will be asked to set-up their 2FA key at their next login, before they can access Chamilo.

This feature might not make it to 1.11.18 and be delayed to 2.0.

PR #5836, opened 2024-10-02
User: Add TOTP 2FA authentication - refs #4431

christianbeeznest:


ywarnier:
Lang vars:
- 'Enter TOTP code' (v)
- 'Submit TOTP' (v)
- '2FA disabled successfully.'
- 'Enable Two-Factor Authentication (2FA)'
- 'Update Settings'
- 'Scan the QR Code to enable 2FA'

ywarnier:
This is great *but*... :-)

Not sure how that works, but the 2FA authentication should send the name of the site (as in the "E-learning portal name" setting in Chamilo, because currently we only send the e-mail address and that doesn't allow someone to remember what service it was for. It should be something like 'Test portal: y@myemail.com' instead of *just* 'y@myemail.com'.

Also, when you setup 2FA, you normally have to confirm the 2FA activation by entering directly the code you get in your 2FA application. You scan the QR code and immediately give the code that appears, and *then* your 2FA is enabled.

Also, when I login and if the code was wrong, I expect a message telling me so, and in my tests I just get a login form refresh but nothing tells me what was wrong.

Finally, to disable 2FA or change your password, you *have to* enter your 2FA code (not just to login).

ywarnier:
I've created #6162 for those improvements, as they don't need any database change (so they can move to the Beta milestone).

PR #5903, opened 2024-11-03
Prevent Cross Site Scripting on new_ticket.php

LudiscapeXApi:
Prevent Cross Site Scripting

PR #5904, opened 2024-11-03
Prevent cross site scripting new_ticket.php

LudiscapeXApi:
Prevent cross site scripting

### DIFF
--- a/public/main/ticket/new_ticket.php
+++ b/public/main/ticket/new_ticket.php
@@ -24,7 +24,7 @@ function updateCourseList(sessionId) {
     }, function (courseList) {
         $("<option>", {
             value: 0,
-            text: "'.get_lang('Select').'"
+            text: '.json_encode(get_lang('Select')).'
         }).appendTo($selectCourse);
 
         if (courseList.length > 0) {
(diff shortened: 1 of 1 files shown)


public/main/ticket/new_ticket.php
```diff
@@ -24,7 +24,7 @@ function updateCourseList(sessionId) {
     }, function (courseList) {
         $("<option>", {
             value: 0,
-            text: "'.get_lang('Select').'"
```
ywarnier, line 27: @LudiscapeXApi I don't understand the purpose of the call to `json_encode()` here. Is that a fix to an issue with the French translation or something like that? Because with json_encode() without second parameter, it seems to me like this would just return the same string with brackes (`{}`) around it...

ywarnier: Answered by @LudiscapeXApi but for some reason not appearing here.
json_encode() transforms (by default, without the specific filtering options) of type `"Sélectionnez le \"barrage\" dès l'ouverture"` into something like `""S\u00e9lectionnez le \"barrage\" d\u00e8s l'ouverture""`, which might effectively prevent it from breaking quoting. However, json_encode() adds quotes around the given string, which is already done by the Chamilo code before this PR, and our preferred method of escaping is to simply uses https://www.php.net/htmlentities. Considering there is no particular need for the JSON format in this call, I will be altering this PR to use `htmlentities()` unless there is a relevant argument about `json_encode()` being better for this particular purpose.

ywarnier: As explained here https://github.com/chamilo/chamilo-lms/wiki/Coding-conventions#security-filter-user-input-that-has-a-chance-to-get-on-screen, filtering of data for XSS should be done *at output time*. This is because altering strings can change them in unforeseen ways, which could change the results in unwanted ways. Instead, XSS should be filtered *when showing the string*, not when adding it to the database (XSS is harmless for a database, but it is harmful when opened in a browser or an e-mail).

For example, in this case, `Security::remove_XSS()` should in fact be called in `TicketManager::sendNotification() `when forming the `$titleEmail` variable and not when inserting it into the database (this is a valid potential vulnerability and I'm sending a patch for this now).

For the `$subject` part, one can see it is already filtered by `Security::remove_XSS()` in `getTicketsByCurrentUser()`.

Did you have a specific observed security issue about this, or are you simply trying to add a filter because you saw there was no filter applied?

ywarnier:
Related to #5903

PR #6016, opened 2025-01-07
Language: Use parent ISO code + self ID for new sub languages

AngelFQC:
refs BT#21568


ywarnier:
Code reviewed and approved

PR #6017, opened 2025-01-07
Internal: Fix HTML resource links for migrated OFAJ portal" - refs BT#22199

christianbeeznest:


ywarnier:
This was a purpose-specific PR created to solve an intermediary issue with an installation in alpha 1

Issue #2043, opened 2017-06-22
Deleting sessions completely


ywarnier:
### Current behavior
When deleting a session, the SessionManager::delete() method is used. However, this method *only* deletes session registers from access_url*, extra_field* and session_rel* tables. There is no management of the track_e_* tables nor all the c_* tables.
Deleting a session should do a thorough clean-up to avoid leaving resources hanging.
To find the list of resources, do a --no-data dump of a Chamilo database and look for all *session_id or id_session fields. This below is a list from v1.9:
```
    $tables = [
        'track_course_ranking' => 'session_id',
        'track_e_access' => 'access_session_id',
        'track_e_attempt' => 'session_id',
        'track_e_attempt_recording' => 'session_id',
        'track_e_exercices' => 'session_id',
        'track_e_course_access' => 'session_id',
        'track_e_default' => 'session_id',
        'track_e_downloads' => 'down_session_id',
        'track_e_item_property' => 'session_id',
        'track_e_lastaccess' => 'access_session_id',
        'track_e_links' => 'links_session_id',
        'track_e_online' => 'session_id',
        'track_e_teacher_in_out' => 'session_id',
        'track_e_uploads' => 'upload_session_id',
        'c_item_property' => 'id_session',
        'c_announcement' => 'session_id',
        'c_attendance' => 'session_id',
        'c_blog' => 'session_id',
        'c_calendar_event' => 'session_id',
        'c_chat_connected' => 'session_id',
        'c_course_description' => 'session_id',
        'c_document' => 'session_id',
        'c_dropbox_file' => 'session_id',
        'c_dropbox_post' => 'session_id',
        'c_dropbox_category' => 'session_id',
        'c_forum_thread_qualify_log' => 'session_id',
        'c_forum_thread_qualify' => 'session_id',
        'c_forum_thread' => 'session_id',
        'c_forum_forum' => 'session_id',
        'c_forum_category' => 'session_id',
        'c_glossary' => 'session_id',
        'c_group_info' => 'session_id',
        'c_link' => 'session_id',
        'c_link_category' => 'session_id',
        'c_lp_view' => 'session_id',
        'c_lp' => 'session_id',
        'c_notebook' => 'session_id',
        'c_quiz_order' => 'session_id',
        'c_quiz' => 'session_id',
        'c_student_publication' => 'session_id',
        'c_survey_invitation' => 'session_id',
        'c_survey' => 'session_id',
        'c_thematic' => 'session_id',
[...]

PR #6038, opened 2025-01-18
Session: Ensure complete session data cleanup with cascade and explicit deletes - refs #2043

christianbeeznest:


ywarnier:
The code looks good and we still have time to detect any issue in further testing before stable

PR #6055, opened 2025-01-28
Group: fix new group category creation

eristich:
For platform configuration with : Allow group category
- Correct bug on new group category creation 
- When you go to the Group page (group.php), if no Group Category exists, it create a new Group Category named by default. 
- Add functionality to prevent from removing Group Category if there is only one.


ywarnier:
Hi @eristich 

Thanks for your contribution and sorry for the delay checking them.
Your PR is good, no real observation that would block me from integrating it.
A few observations to improve later contributions:
- when updating the headers of an old function/method, don't hesitate to complete it with the right type hinting. As Chamilo 2.0 will only support PHP 8.2 and above, we try to be more strict about the declaration of functions/methods. In this case, you provided default values for method `create_category()` but didn't provide any required type to the data provided, which would have been a nice addition.
- the PHPDoc block of that same method is outdated, so don't hesitate to update it with the new params. If you use a student license of PHPStorm, for example, and set it to check for Symfony style code conventions, it will pretty much suggest and complete the header almost by itself.
- if you ever spot `get_lang()` calls with CamelCase writing, like `get_lang('AddCategory')` in `group.php`, this comes from a bad conversion of language variables that will prevent their translation. Look into translations/messages.pot for similar strings without CamelCase, like `Add category` in this case, and replace in the `get_lang()` call in a separate commit (because this is a visual change, not a logic change, so no logic impact for automated testing), like this: `get_lang('Add category')`. This will help us progress faster on releasing a great 2.0 stable

ywarnier:
You commit message was good too, by the way :+1:

eristich:
Hi, no problem. Thank you for taking the time to read and comment this pull request. I'll take your recommendations in account for future modifications 🔥.

Issue #6045, opened 2025-01-24
[C2] Remove user.registration_date


ywarnier:
The `user` table still maintains a `registration_date` field which is a functional duplicate of `user.created_at`.

- [x] Remove `registration_date` in the entity.
- [x] add a migration to deal with the field from 1.11.x, verifying that an empty `created_at` will be filled by the `registration_date` value before deletion.
- [x] verify (including in backup/restore code) that this doesn't cause issues with user imports, user exports, cronjobs or other places where `registration_date` might be used. In such cases, just replace by `created_at`.

christianbeeznest:
It is implemented in this PR https://github.com/chamilo/chamilo-lms/pull/6059

Thanks for confirmation.

ywarnier:
All good for me

PR #6059, opened 2025-01-30
User: Remove user.registration_date #6045

christianbeeznest:


christianbeeznest:
It is implemented in this PR https://github.com/chamilo/chamilo-lms/pull/6059

Thanks for confirmation.

PR #6062, opened 2025-01-31
User: add advanced user edition with bulk and ajax updates - refs BT#22305

christianbeeznest:


main/admin/user_advanced_edit.php
```diff
@@ -0,0 +1,330 @@
...
+});
+</script>';
+
+$formContent = $form->returnForm();
+
+// Render page
+$tpl = new Template($tool_name);
+$tpl->assign('actions', $toolbarActions);
+$tpl->assign('message', $message);
+$tpl->assign('content', $formContent . $fieldSelector . $tableResult . (!empty($users) ? '<button class="btn btn-success" id="saveAll">'.get_lang('SaveAll').'</button>' : ''));
```
ywarnier, line 329: "SaveAll" could be replace (in get_lang()) by SaveChanges, to avoid introducing a new variable.

PR #6079, opened 2025-02-09
Internal: Add security fixes from 1.11.x

christianbeeznest:


public/main/inc/lib/formvalidator/FormValidator.class.php
```diff
@@ -1145,6 +1148,17 @@ public function addHtmlEditor(
...
 
+    /**
+     * Prevent execution of event handlers in HTML elements.
+     *
+     * @param string $html
+     * @return string
+     */
+    function attr_on_filter($html) {
+        $prefix = uniqid('data-cke-').'-';
+        return preg_replace('/(\s)(on)/i', '$1'.$prefix.'$2', $html);
```
ywarnier, line 1159: @christianbeeznest This must be updated with the new form in 1.11.x as it was detected that this pattern here did not work
See 8c4e6436

ywarnier (commented):
One issue to review

ywarnier:
The rest seems fine. We'll see when the first pentest is over.

PR #6080, opened 2025-02-11
User: Add extra fields filter to advanced user search - refs BT#22305

christianbeeznest:


main/admin/user_advanced_edit.php
```diff
@@ -116,7 +125,12 @@
     }
     unset($user);
```
ywarnier, line 126: Given this `unset($user);`, I'm not sure the `$user['extra_fields'] = $formattedExtraFields;` is useful.

PR #6092, opened 2025-02-18
Pass/fail conditional session course subscription - refs BT#22403

AngelFQC:


ywarnier:
Tested and approved.
The text appearing could benefit from some styling ('alert' if failed, 'success' if succeeded), but I see from the code that it would change the behaviour of where the styling is defined, which is not entirely desirable code-wise.

Issue #2675, opened 2018-09-18
Improve plugin installation, add enable/disable 


jmontoyaa:
### Current behavior / Resultado actual / Résultat actuel

Improve UI to plugin installation:

- Remove checkbox and mass action list
- Add install/uninstall button
- Add enable/disable button (doesn't delete all plugin table)
- Modify "hook" to **not fire** events for disabled plugins
- Do not parse disabled plugins in regions

PR #6116, opened 2025-03-06
Internal: Improve plugin management UI & actions - refs #2675

christianbeeznest:


AngelFQC:
Changes here will be included in the PR #6127

PR #6117, opened 2025-03-06
Internal: Add direct lessons access button & hide header in reduced mode - refs BT#22451

christianbeeznest:


ywarnier:
The code looks good. Not sure about the necessity of the list.tpl eventListener, but apart from that, all good.
The feature also works well, apart from the fact that the alt text is not translated, which I will fix.

PR #6130, opened 2025-03-12
Exercise: Add validation to enforce correct answer and positive score - refs BT#22404

christianbeeznest:


ywarnier:
Try to keep the number of language variables to the minimum. In this case, AtLeastOneCorrectAnswerRequired ~= MustSelectCorrectAnswer and AtLeastOneCorrectAnswerMustHavePositiveWeighting ~= CorrectAnswerMustHavePositiveScore.
More language variables mean more time for translators, and that means a smaller ratio of translator contributions in the long run.

ywarnier:
> Try to keep the number of language variables to the minimum. In this case, AtLeastOneCorrectAnswerRequired ~= MustSelectCorrectAnswer and AtLeastOneCorrectAnswerMustHavePositiveWeighting ~= CorrectAnswerMustHavePositiveScore. More language variables mean more time for translators, and that means a smaller ratio of translator contributions in the long run.

Nevermind, I just realized it is because of the difference between single answer and multiple answer ! :man_facepalming:

Issue #6154, opened 2025-03-20
[C2] Add feature to select visibility of courses and sessions in the catalogue, including per URL


ywarnier:
A number of users have asked for a feature to help define what should appear in the course (or session) catalogue. But having this feature also implies making it compatible with multiple access URLs.

This means that such a relationship could only be expressed through tables that include the course (or session) *and* the URL.

Some users have also asked for specific courses (or sessions) to be available in the catalogue but only for a specific group of users (as in the usergroup table in Chamilo).

- [x] Structure

To enable this, the following tables should be created:

```
catalogue_course_rel_access_url_rel_usergroup
  id int PRIMARY KEY,
  course_id int,
  access_url_id int,
  usergroup_id int DEFAULT NULL

catalogue_session_rel_access_url_rel_usergroup
  id int PRIMARY KEY,
  session_id int,
  access_url_id int,
  usergroup_id int DEFAUT NULL
```

- [x] Filter rules

Based on this structure, we can consider that:
- if the tables are empty for the current URL, all courses and sessions can be shown (if the catalogues are enabled)
- if just *one* course or session appears for the current URL, then the catalogue should only show those courses or sessions that are there
- if the usergroup field is NULL, then the course or session can appear to all users in that URL
- if the usegroup field is NOT NULL, then the corresponding course or session only appears to users that belong to that usergroup
- one same course (or session) can be assigned to *several usergroups* in the same URL. In this case, it appears only once but for any user of any of those usergroups

- [x] Migration

Because Chamilo 1.* doesn't have this, the tables should be empty when migrating, so all courses will appear for all URLs. Users can then decide which courses or sessions to add to the catalogue by clicking an icon in the admin courses list (use `ObjectIcon::CATALOGUE`) (they are all disabled by default).

However, taking into account that an option to use the extra field `show_in_catalogue` existed in Chamilo 1.*, the migration should look for that field and insert a record in `catalogue_course_rel_access_url_rel_usergroup` for the current URL. Something like `SELECT fv.item_id FROM extra_field_values fv, extra_field f WHERE f.item_type = 2 AND f.variable = 'show_in_catalogue' AND f.id = fv.field_id AND fv.field_value = 1` to feed the list of course IDs to your algorithm.

Replaces #5990

christianbeeznest:
It is implemented with this PR https://github.com/chamilo/chamilo-lms/pull/6159

Thanks for confirmation.

ywarnier:
[...]

PR #6159, opened 2025-03-21
Internal: Implement catalogue filters and toggle icon for courses - refs #6154

christianbeeznest:


ywarnier:
The code looks good. I'm having an issue with the "My courses" list at the moment but I don't think it's related.
I also have an issue to create a session.
I tested it from the courses list and it works.
I was expecting a way to assign a course to the catalogue for a specific usergroup but haven't found an icon or button to do that yet.
I'm logging a new issue for that part.

Issue #6153, opened 2025-03-20
[C2] Update BBB to use new database structure (videoconference tables)


ywarnier:

christianbeeznest:
This PR https://github.com/chamilo/chamilo-lms/issues/6153 migrates the BigBlueButton (BBB) plugin in Chamilo 2 to use the new videoconference database structure, replacing all direct SQL access (Database::select, insert, update, delete) with Doctrine-based repositories (ConferenceMeeting, ConferenceActivity, ConferenceRecording).

All data interactions now use Symfony and Doctrine ORM.

Note: This refactor requires thorough testing with a live BigBlueButton server to ensure meeting creation, joining, recording, and deletion are fully functional.

NicoDucou:
Hi @christianbeeznest 

I've put a few comment on things to correct in your PR for it to work. I let you revise my comment and implement the proposed solution or a better one.
I also indicate here that on the listing.php page when clicking on the delete meeting icon it does not work, it is still present in the DB and there is no message on the interface indicating the success or the error.
It works when putting a meeting invisible or visible.
I did not check anything related to recording because I can not activate it.

christianbeeznest:
> Hi [@christianbeeznest](https://github.com/christianbeeznest)
> 
> I've put a few comment on things to correct in your PR for it to work. I let you revise my comment and implement the proposed solution or a better one. I also indicate here that on the listing.php page when clicking on the delete meeting icon it does not work, it is still present in the DB and there is no message on the interface indicating the success or the error. It works when putting a meeting invisible or visible. I did not check anything related to recording because I can not activate it.

They are checked and fixed in the next commits of the same PR . 

Thanks for confirmation.

NicoDucou:
I've tested again, and most of it works now.
I've put a new comment because there is still an error 500 when coming back from the visio.
The delete action is also still not working.

christianbeeznest:
> I've tested again, and most of it works now. I've put a new comment because there is still an error 500 when coming back from the visio. The delete action is also still not working.

I just tested in ofajmigdev , delete action is working now, with the last changes.

Thanks for confirmation.

NicoDucou:
I all seems to work fine with the general use of BBB meetings.
I did not test the recording because we need to be able to configure it and it depends on the general plugin installation system that is not ready yet to be able to access the configuration of the plugin.
I do not know if I close the issue and create a new one for that or if we leave this issue here until we test this part.
@ywarnier what do you think ?

[...]

PR #6161, opened 2025-03-22
Plugin: Refactor BBB to use new conference tables - refs #6153

christianbeeznest:


public/plugin/bbb/lib/bbb.lib.php
```diff
@@ -747,55 +761,51 @@ public function getMeetingParticipantInfo($meetingId, $userId)
...
-                        'plugin_bbb_room',
-                        ['out_at' => api_get_utc_datetime(), 'close' => BBBPlugin::ROOM_CLOSE],
-                        ['id = ? ' => $roomId]
-                    );
-                }
+        /** @var ConferenceActivityRepository $repo */
+        $repo = $em->getRepository(ConferenceActivity::class);
+
+        $meeting = $em->getRepository(ConferenceMeeting::class)->find($meetingId);
+        $user = $em->getRepository(User::class)->find($participantId);
```
NicoDucou, line 772: This returns an error and I do not know why.
`PHP Fatal error:  Uncaught Doctrine\\Persistence\\Mapping\\MappingException: Class 'Chamilo\\UserBundle\\Entity\\User' does not exist in /var/www/ofajmigdev.beeznest.com/www/vendor/doctrine/persistence/src/Persistence/Mapping/MappingException.php:80\nStack trace:\n#0`

To make it work I changed it with  :
`$user = api_get_user_entity($participantId);`

But I'm not sure it's a good idea to use a legacy function, it would be better to make it work with the repo.

public/plugin/bbb/lib/bbb.lib.php
```diff
@@ -1898,16 +1884,21 @@ public function getConferenceUrl()
...
-            'plugin_bbb_meeting',
-            array('where' => array('meeting_name = ? AND status = 1 ' => $name)),
-            'first'
-        );
+        $em = Database::getManager();
+        /** @var ConferenceMeetingRepository $repo */
+        $repo = $em->getRepository(ConferenceMeeting::class);
+
+        $qb = $repo->createQueryBuilder('m')
+            ->where('m.meetingName = :name')
```
NicoDucou, line 1894: The meetingName column does not exist anymore it is title now so it sould be changed like this :
->where('m.title = :name')

public/plugin/bbb/listing.php
```diff
@@ -182,12 +188,7 @@
             }
 
             $remoteId = Database::escape_string($_GET['remote_id']);
-            $meetingData = Database::select(
-                '*',
-                Database::get_main_table('plugin_bbb_meeting'),
-                ['where' => ['remote_id = ? AND access_url = ?' => [$remoteId, api_get_current_access_url_id()]]],
-                'first'
-            );
+            $meetingData = $meetingRepository->findOneByRemoteIdAndAccessUrl($remoteId, api_get_current_access_url_id());
```
NicoDucou, line 192: This do not return the values of the column user_id which is needed on line 199 and generated an error.
We should get an object user but I do not find it to be able to adapt line 199 so this still needs to be fixed.

public/plugin/bbb/lib/bbb.lib.php
```diff
@@ -913,225 +953,178 @@ public function getMeetings(
...
-            if (!empty($courseInfo)) {
-                $courseCode = $courseInfo['code'];
+            $dateStart = (new \DateTime($dateRange['search_meeting_start']))->setTime(0, 0, 0);
+            $dateEnd = (new \DateTime($dateRange['search_meeting_end']))->setTime(23, 59, 59);
+            $meetings = $repo->findByDateRange($dateStart, $dateEnd);
+        } elseif ($this->isGlobalConference()) {
+            $meetings = $repo->findBy([
+                'course' => null,
+                'user' => api_get_user_entity($this->userId),
+                'status' => 1,
```
NicoDucou, line 969: To be removed, in BBB we should return all the conferences Opened or Not

public/plugin/bbb/lib/bbb.lib.php
```diff
@@ -913,225 +953,178 @@ public function getMeetings(
...
+                'accessUrl' => api_get_url_entity($this->accessUrl),
+            ]);
+        } elseif ($this->isGlobalConferencePerUserEnabled()) {
+            $meetings = $repo->findBy([
+                'course' => api_get_course_entity($courseId),
+                'session' => api_get_session_entity($sessionId),
+                'user' => api_get_user_entity($this->userId),
+                'status' => 1,
```
NicoDucou, line 977: To be removed, in BBB we should return all the meetings opened or not

public/plugin/bbb/lib/bbb.lib.php
```diff
@@ -913,225 +953,178 @@ public function getMeetings(
...
+                'accessUrl' => api_get_url_entity($this->accessUrl),
+            ]);
+        } else {
+            $criteria = [
+                'course' => api_get_course_entity($courseId),
+                'session' => api_get_session_entity($sessionId),
+                'status' => 1,
```
NicoDucou, line 984: To be removed, in BBB we should return all the meetings opened or not

public/plugin/bbb/listing.php
NicoDucou: There is still a problem on line 196 because when returning from the video conference it does not find the $meetingData['id'] and gives a 500 error :
`PHP Fatal error:  Uncaught ErrorException: Warning: Undefined array key "id" in /var/www/.../public/plugin/bbb/listing.php:196`

PR #6188, opened 2025-04-02
Session: Add Excel export of certified users in course session with extra fields - refs BT#22533

christianbeeznest:


ywarnier:
Code approved. Feature testing pending.

PR #6194, opened 2025-04-03
Ticket: Access URL integration and ticket module refactoring

christianbeeznest:


ywarnier:
Untested, but the code looks fine and we have time for debugging during beta development

Issue #6163, opened 2025-03-24
[C2] Add selectors for courses and sessions to show in catalogue only for given usergroups


ywarnier:
Thanks to #6154, we now have the database structure to make some courses or sessions available only to some usergroups in the catalogue. However, the interface for that doesn't exist yet.

This should go as a tab (called "Catalogue access") in the course edit page and the session edit page, because it will probably be the most logical place to find this configuration. On the page, we should see double multiselect (like to assign a course to a session) showing all usergroups on the left and the selected usergroups on the right.

A text help at the top should say:
"Select classes for which this [course/session] will be visible for subscription in the catalogue. Subscription rules still apply apart from it being visible in the catalogue."

_Originally posted by @ywarnier in https://github.com/chamilo/chamilo-lms/issues/6159#issuecomment-2748431435_

christianbeeznest:
It is implemented with this PR https://github.com/chamilo/chamilo-lms/pull/6204

Thanks for confirmation.

ywarnier:
Thanks!

PR #6204, opened 2025-04-06
Catalogue: Add "Catalogue access" tab to restrict visibility by usergroup - refs #6163

christianbeeznest:


ywarnier:
Code slightly modified to add course title and session title in tabs (otherwise difficult to guess).
Tested forms (not tested end result in catalogue visibility).
Works for me. Merging.

Issue #4814, opened 2023-07-17
Send announcement to a class


riesaau:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I need to send an announcement to a large number of users who are subscribed to a course who i have put in a new class, but I don't want to send it to all of the others who are not in that class.

**Describe the solution you'd like**
I want to be able to send an announcement to a class.  There may be a way of doing this already but I can't find it.

**Additional context**
![Screenshot_2023-07-17_at_18_03_12](https://github.com/chamilo/chamilo-lms/assets/97856892/94e12710-220a-4458-819c-30acdb356670)

riesaau:
It's not an issue if it is a very small class, as I can just search for the people one by one and add them to the announcement list, but if it is a lot of students it would be good to add them to the announcement field by class or by csv file.

ywarnier:
Hi @riesaau sorry for the late reply. Your suggestion is good and we will include it in a future version, but it is a bit too much to add it to 1.11.22 coming out this week.

christianbeeznest:
It is implemented in this PR https://github.com/chamilo/chamilo-lms/pull/6214

Thanks for confirmation.

christianbeeznest:
It is implemented to send announcements to users of session classes in this commit https://github.com/chamilo/chamilo-lms/pull/6214/commits/1adb8eeb10626cacf0ab9d41bb1dd2a440577619

When in session context, all classes linked to the session are listed, and all their users are included, even if not enrolled in the course.

Thanks for confirmation.

ywarnier:
This has been implemented and will be available in Chamilo 2.0

PR #6214, opened 2025-04-09
Announcement: Add support to send announcement to a class - refs #4814

christianbeeznest:


ywarnier:
Code approved.
User-tested.
Good contrib!

PR #6222, opened 2025-04-12
Learnpath: Add support for selective LP item PDF export - refs #2969

christianbeeznest:


ywarnier:
Code looks good. Feature untested.

Issue #5634, opened 2024-07-04
[C2] Add "duration" type extra field


ywarnier:
Add a new extra field type `duration` that is expressed in seconds and shows in `hh:ii:ss` format.

More and more requirements have appeared recently to be able to give an estimate study/completion time for any item. Adding the availabillity of such a field type would make it easier to theme those times everywhere necessary.

This would probably be a new extra_field.value_type (`FIELD_TYPE_DURATION`in extra_field.lib.php) of 28.

christianbeeznest:
It is added with this PR https://github.com/chamilo/chamilo-lms/pull/6238

Thanks for confirmation.

ywarnier:
That works well. There's another issue with me not being able to change "changeable" fields in my profile page, but that's for another issue.

PR #6238, opened 2025-04-17
Extrafield: Support duration extra field (hh:mm:ss) - refs #5634

christianbeeznest:


ywarnier:
Code looks good.
Lang variable to be added (by me).
User testing still required.

Issue #4454, opened 2022-11-10
Warn when deleting a document that is used by a learning path


ywarnier:
Currently (1.11.16), it is possible to delete a document from the documents tool even if the document is actively used in a learning path.

Add a visual warning and a confirmation button when trying to delete a document and finding it is in use in the c_lp_item.path table for the same course. The warning should be something like:

$DocumentDeletionWarningUsedInLP = "The document (or one of the documents) you are trying to delete is currently used (at least) in learning path %s. Deleting it will create an inconsistency to your learners by showing a 'Page not found' message instead of the document wherever it was used in the learning path. Instead, you could edit the document or use the 'Replace document' icon to replace it with a white page or an explanation about why it has been removed. Are you sure you want to delete it?"

christianbeeznest:
It is implemented with this PR https://github.com/chamilo/chamilo-lms/pull/6239

Thanks for confirmation.

ywarnier:
<img width="2177" height="907" alt="Image" src="https://github.com/user-attachments/assets/b2b9c5e4-494c-4643-a58b-69c043077f5f" />
Works perfectly. Thanks!

PR #6239, opened 2025-04-18
Document: Add LP usage warning when deleting documents - refs #4454

christianbeeznest:


ywarnier:
Code looks good.
Note to self : some language variables to add.
Still requires user testing.

Issue #5454, opened 2024-04-29
[C2] Pages: add slug edition, return to list and pagination memory


ywarnier:
In the new "Pages" feature in administration, add a few improvements:
- [x] Add a field to edit the slug (called "Friendly URL")
- [x] Load pages using the slug (instead of the title) when the page has enabled=1
- [x] Show a link to preview the page using the page id and "/preview" only for admins
- [x] In the edit page, add a button to go "Back" to the list of pages
- [x] In the list of pages, if I previoously selected pagination of 50 items per page, keep that setting when I go to edit and return

christianbeeznest:
It is implemented in this PR https://github.com/chamilo/chamilo-lms/pull/6242/commits/916de393bfb6588ee61982b49dfb82207b2742e0

Thanks for confirmation.

christianbeeznest:
It is implemented the pending points in this PR https://github.com/chamilo/chamilo-lms/pull/6308 merged.

ywarnier:
@christianbeeznest 
Preview does not work.
I'm editing https://testing25.beeznest.com/resources/pages/edit?id=/api/pages/5&page=1&itemsPerPage=5 and clicking the preview link that says https://testing25.beeznest.com/pages/5/preview, which gives me a NotFoundHttpException.

christianbeeznest:
It is fixed with this commit https://github.com/chamilo/chamilo-lms/pull/6658/commits/3eedae72bc3f263b12296a31a04eaf0e1fabb48a

Thanks for confirmation.

ywarnier:
The preview doesn't seem to take all styles with it, and it only previews the previously saved version (if I edit something below, it doesn't appear in the preview until I save the page), but at least it works to preview.

Here a (confusing for my use of colors that do not match the letters, sorry) example of not previewing with colors:
<img width="944" height="540" alt="Image" src="https://github.com/user-attachments/assets/c5bda480-2e0f-4ae7-a155-50353a064f82" />

The issues mentioned have all been fixed or implemented, so I'm closing, but you might want to check that preview issue.

PR #6242, opened 2025-04-20
Pages: Add slug edition, return to list and pagination memory - refs #5454

christianbeeznest:


ywarnier:
Code looks good.
Language variables to be added (by me).
Requires user testing.

Issue #5840, opened 2024-10-04
[C2] Control changes in roles


NicoDucou:
When changing a user role we should control if it is compatible with the assignation this user already have and show a message to indicate the problems that would happen.

For example if a user has a teacher role and we assign it a course A in which he is the teacher and we assign him as general coach of a session B. If we try to change the role of this user from teacher to student it should not allow us and show a message indicating the the student role is incompatible with the assignation as a teacher in course A and the assignation as general manager in session B and indicate that this should be change to be able to change the user role.

This is a specific example but we should verify the compatiblity for every change.

To be implemented in Chamilo 2

christianbeeznest:
Hi @NicoDucou ,

It is implemented with this PR https://github.com/chamilo/chamilo-lms/pull/6244

Thanks for confirmation.

ywarnier:
Done, with a minor issue registered as #6461

PR #6244, opened 2025-04-20
User: Control changes in roles - refs #5840

christianbeeznest:


ywarnier:
Code looks good (although not sure that                 `$courseTeacherCount = $userObj->getCourses()->count();` actually gets the courses where the user is a teacher (as opposed to a student).
Lang variables to be added (by me).
User testing still missing.

Issue #6105, opened 2025-02-28
[C2] "Last connexion" column empty in users list as admin


ywarnier:
**Describe**
As an admin, when I look at the users list, the "Last connexion" column is empty.
The registration date is not empty, so all good on that side.
I logged in as the first user in the screenshot, and you can see the field is empty.

**To Reproduce**
1. Go to '/main/admin/user_list.php' as admin
2. See that the Last connexion column is empty for all users

**Expected behavior**
A date similar to the one in "Registration date" should show in the "Last connexion" column.

**Screenshots**

![Image](https://github.com/user-attachments/assets/97c86079-554c-4afb-9989-45d6f451ceef)

**Desktop (please complete the following information):**
Not important

**Server
 - OS: Linux Ubuntu
 - Version of Chamilo:  2.0 from git Monday
 - Version of PHP: 8.3

ywarnier:
Fixed.

PR #6262, opened 2025-05-02
Internal: fix bug that prevent display last user connection - refs #6105

eristich:


ywarnier:
@eristich Where is the problem showing when this PR is *not* applied?
I try to reproduce it, but as far as I can see, everytime I log in, my user.last_login field is updated and I have a new record in track_e_login.
However, I see that track_e_access is not updated for every tool (documents is missing), but this seems to be unrelated.

ywarnier:
No answer from contributer. Closing

PR #6276, opened 2025-05-06
Upgrade primevue to v4

AngelFQC:


NicoDucou:
Tried in course in various tools and in administration pages and in messages and all seems to work fine so I accept the PR.

Issue #3498, opened 2020-09-05
BigBlueButton always visible


MAMEK70:
**Describe**
BigBlueButton sessions URL is always visible even when the "Global conference link only visible for these user roles" option is set only to Administrator

**To Reproduce**
1. Go to 'adminstration/plugins/bbb/'
2. Scroll down to "Global conference link only visible for these user roles" and set it to "Administrator"
3. Go to a Course, open a Videoconference
4. The error: The VideoConference URL is visible to all (teacher, student, etc...)

**Expected behavior**
The videoconference url should be hidden to students and teachers (in this case)

**Screenshots**
![image](https://user-images.githubusercontent.com/70807894/92306479-9a852400-ef8f-11ea-8dd4-c5b41eba4f45.png)
![image](https://user-images.githubusercontent.com/70807894/92306490-b7b9f280-ef8f-11ea-8374-f51cfe8772fb.png)

**Desktop**
 - Windows 10
 - chrome 85.0.4183.83 (Official Build) (64-bit)

**Server
 - Version of Chamilo:  1.11.12

aelalfey:
@MAMEK70 

I have the same issue and more. Check this https://github.com/chamilo/chamilo-lms/issues/3472

MAMEK70:
@aelalfey 
Pls let me know if you find a solution...

jmontoyaa:
The global conference link it's located here ( see screenshot link Videoconference below "Edit profile").

In your screenshot your're showing a course videoconference.

![Screenshot_20200914_093457](https://user-images.githubusercontent.com/158935/93057281-ef026080-f66d-11ea-8182-2c9078913d8b.png)

lonesomewalker:
@jmontoyaa 
Yes, but this also should be possible to disable by default ;-)

ywarnier:
So just to make this issue clear: the request is to add an option to avoid showing the *link* (in text) to the videoconference to the users, because in some cases we just want them to see in an obvious way the server on which it is hosted (or something like that).

ywarnier:
This requires adding a configuration option for the videoconference, so we need this pre-alpha3 @christianbeeznest

christianbeeznest:
It is implemented logic to restrict access to global videoconference link based on global_conference_allow_roles setting.
No DB changes required — setting already exists in plugin configuration.
Ready for testing before alpha3.

ywarnier:
The requirement is to add an option to hide the link block in the course tool for videoconference

[...]

PR #6336, opened 2025-06-04
Plugin: Add config-based control for global conference roles visibility in BBB - refs #3498

christianbeeznest:


ywarnier:
Looks good. Merging.

Issue #5952, opened 2024-12-04
[C2] Course button "Hide all" do not hide all the tools (few missing tools


NicoDucou:
In the homepage of a course, if we click on "Hide all" (page like type https://xxx/course/3/home).
We login the page again some tools are still marked as visible : Chat, Notebook and Wiki.

ywarnier:
As far as I can test, this works correctly now.

PR #6373, opened 2025-06-19
Fix bug #5952  Course button "Hide all" do not hide

chamilo-uga:


ywarnier:
Seems logical to me.
Note: those 3 tools are not yet working in C2.

Issue #3819, opened 2021-03-08
Add "validation" server type to configuration settings


ywarnier:
In 1.11.x, on the first page of platform settings, the "Server type" allows the admin to choose between "production" and "test". However, we need a 3rd type.

 - Production: as is now (no change)
 - Validation/Testing: as production (with maybe minor changes), but does not register stats to chamilo.org
 - Development: as current "Test server" type (shows debug information)

This will be useful to be more consistent in reporting replica portals to the Chamilo.org (through version.chamilo.org) statistics engine (these should not be counted, or counted in a separate count).

christianbeeznest:
It is done with this PR https://github.com/chamilo/chamilo-lms/pull/6376

Thanks for confirmation.

ywarnier:
Looks good to me!

PR #6376, opened 2025-06-19
Internal: Add "validation" server type to configuration settings - refs #3819

christianbeeznest:


public/main/inc/lib/api.lib.php
```diff
@@ -2712,10 +2712,16 @@ function api_get_setting($variable, $isArray = false, $key = null)
             if (in_array($environment, $test)) {
```
ywarnier: I think that logic needs to change, because if the environment (I suppose from APP_ENV in .env) is "dev", then `server_type` can never be "prod".
I think those (APP_ENV and server_type) are independent

ywarnier: Confirmed. Please ignore the $environment to set the current server_type.

ywarnier (changes requested):
A change in logic is required. Only check the value of settings_current, with a default value of 'prof'.

christianbeeznest:
Thank you, it is done in the next commit

ywarnier:
Merged. Try to avoid mixing with something unrelated ($full/$short change), but otherwise OK.

Issue #3143, opened 2020-03-27
Deleting course does not delete BBB conference recordings


ywarnier:
When the videoconference plugin (BBB) is enabled, course recordings can be created.
When a course is deleted, there is currently no process to delete the recording, so recordings remain there, unattached.

Add an option to the BBB plugin to delete recordings on course or session deletion.
Use a hook to delete them through the bbb::deleteRecording() method.

christianbeeznest:
It is implemented with this PR https://github.com/chamilo/chamilo-lms/pull/6377

Requires more testing.

PR #6377, opened 2025-06-20
Plugin: Cleanup BBB recordings on course/session deletion - refs #3143

christianbeeznest:


ywarnier:
I cannot really test it for now (we can still do that in beta) but the code looks good. Closing.

Issue #2512, opened 2018-04-30
Add automatic switch to next step in learning path, only if video


ywarnier:
### Current behavior
If a learning path is made almost exclusively of videos, the course stops at the end of each video.

### Expected behavior
Be able to "let it run", so that, at the end of each video, the next step is started automatically (after 10 seconds).

### Steps to develop
 - consider video files as a new type of document (file, folder, video)
 - when adding a video (file) as a step in a learning path, register this step at lp_item.type = video
 - when showing this item, intercept the lp_view.php and add the corresponding video player tags (so it is not necessary to embed the video in an HTML page)
 - modify the media player with a callback to "switch_item()" of the learning path, so that the next item is automatically loaded 10 seconds after the end of the video
 - optional: add an option in the lp table to enable/disable auto-forward for videos

Extra: in the table of contents of the learning path, show the time [hh:mm:ss] of the video (register that as an extra_field of the document, maybe?)

ywarnier:
If this is not possible in a realistic timeframe, then using the postroll plugin to show a button to move to the next step would be nice also: https://github.com/mediaelement/mediaelement-plugins/blob/master/docs/postroll.md
The playlist plugin also seems to be a possibility, as it allows one to execute an action to pass to the next element.

christianbeeznest:
It is implemented with this PR https://github.com/chamilo/chamilo-lms/pull/6419 , it requires more testing.

Thanks for confirmation.

ywarnier:
Tested and approved.

PR #6419, opened 2025-06-30
Learnpath: Add automatic switch to next step in learning path, only if video - refs #2512

christianbeeznest:


ywarnier (commented):
Code generally looks good overall.
I would prefer declaring and using a new constant "TOOL_DOC_VIDEO" rather than the string "video" in learnpath.class.php an others.
Sending you PR with a conversion of quotes to double quotes make it much harder to review.
New potential language terms:
- 'Videos'
- 'No videos found'

ywarnier:
Merging to get the DB structure for Alpha 3.
Feature remains to be tested.

Issue #6254, opened 2025-04-24
[C2] Convert file uploads/downloads to OneUpFlySystem


ywarnier:
This includes reviewing and potentially replacing all call to (non exhaustive list):
- [ ] basename()
- [ ] chmod()
- [ ] chown()
- [ ] closedir()
- [ ] connection_status()
- [ ] copy()
- [ ] dir()
- [ ] fclose()
- [ ] feof()
- [ ] file_exists()
- [ ] file_get_contents()
- [ ] filemtime()
- [ ] file_put_contents()
- [ ] filesize()
- [ ] uses of class Fllesystem()
- [ ] fopen()
- [ ] fpassthru()
- [ ] fread()
- [ ] fseek()
- [ ] fwrite()
- [ ] is_dir()
- [ ] is_file()
- [ ] link()
- [ ] local copyr()
- [ ] mime_content_type()
- [ ] mkdir()
- [ ] opendir()
- [ ] pathinfo()
- [ ] readdir()
- [ ] readfile()
- [ ] realpath()
- [ ] tempnam()
- [ ] rmdir() (and local rmdirr)
- [ ] scandir()
- [ ] unlink()
- [ ] uses of class UploadedFile

We should also check all uses of exec() and system() to see if they do not treat a file path directly.
- [ ] exec()
- [ ] system()

christianbeeznest:
Hi @ywarnier,

Thanks for opening this task!

I’ve reviewed the list of functions mentioned in the issue and would like to confirm the scope before proceeding with the migration work.

Below is how I’m classifying them in terms of Flysystem replacement:

## ✅ Review plan for migration to Flysystem

Below is a checklist of functions/classes mentioned in the task, with comments on whether they should be replaced with Flysystem, kept as-is, or analyzed case by case.

| Function / Class         | Action                                                                                          |
|---------------------------|-------------------------------------------------------------------------------------------------|
| basename()               | ✅ **Keep.** Pure string operation. No need to replace.                                          |
| chmod()                  | ⚠️ Usually keep. Flysystem does not manage local filesystem permissions directly.               |
| chown()                  | ⚠️ Usually keep. Not relevant for Flysystem adapters.                                            |
| closedir()               | ✅ Replace with Flysystem `listContents()` if iterating directories.                             |
| connection_status()      | ✅ Keep. Unrelated to filesystem.                                                                |
| copy()                   | ✅ Replace with Flysystem `copy()` or read/write pair.                                           |
[...]

PR #6451, opened 2025-07-09
Internal: Implement Flysystem FileHelper - refs #6254

christianbeeznest:


ywarnier:
Merging to have it in alpha 3. The complete overhaul of file-related functions is still to be done so the related issue remains open, but the FileHelper looks good in the meantime.

Issue #3793, opened 2021-02-23
Upload themes (css)


jmontoyaa:
Allow to upload css themes inside the public/ folder in order to improve performance instead of saving the files in var/

ywarnier:
We need the option to upload new CSS themes to Chamilo 2.0. Otherwise it will be blocking for many.

The things that need to be uploaded:
- replacement for assets/css/app.scss (see pre-processing mechanism here: https://github.com/chamilo/chamilo-lms/wiki/Graphical-design-guide#appcss)
- replacement for the logo at assets/css/themes/chamilo/images/header-logo.svg

Technically, the new styles should all be uploaded in var/ because that's the only folder we want users to be able to write to.
There is currently a var/templates/ folder, but it seems more dedicated to... templates (HTML templates).
I guess we should create a var/css/themes/[theme-name]/ to hold the SCSS file(s).
Whatever the pre-processing mechanism is now to generate CSS from SCSS (when you upload a new CSS theme), it should first load assets/css/app.scss and whatever it loads now, then find the CSS theme configured in settings_current, then load the custom SCSS file(s) from var/css/themes/[theme-name]/ and, with that, generate a /build/css/app.css and a custom /build/css/[theme-name]/app.css that has to be loaded when that style is selected.

It is not necessary to offer the possibility to *update* an existing theme. Each new theme version should have a version number, because this will avoid bad caching processes.
Upload should be done in zip format as these are always multiple-files. Look at 1.11.x implementation for details on filtering.

daniboygg:
Reviewing the pull request https://github.com/chamilo/chamilo-lms/pull/4753 we decide the following steps for making this feature: 

1. Send a JSON with Chamilo colors to the backend
```
{
    "--color-primary-base": "46 117 163",
    "--color-primary-gradient": "156 194 218",
    "--color-secondary-base": "243 126 47",
    "--color-secondary-gradient": "224 100 16",
    "--color-tertiary-base": "0 0 0",
    "--color-tertiary-gradient": "51 51 51",
    "--color-success-base": "119 170 12",
    "--color-success-gradient": "84 119 8",
    "--color-danger-base": "223 59 59"
}
```
3. The backend should store this as a JSON in the database
[...]

PR #6458, opened 2025-07-10
Internal: Add default color theme on install/migrate - refs #3793

ywarnier:
Missing URL assignment and fixing issue on use of array for "variables".


ywarnier:
Substituted by #6466

PR #6485, opened 2025-07-19
Show four columns vs three actually in course grid

yverhenne:


.gitignore
```diff
@@ -16,6 +16,7 @@ app/logs/*
 /app/config/mail.conf.php
 /app/config/profile.conf.php
 /app/config/configuration.php
+composer.lock
```
yverhenne, line 19: Cette ligne n'est pas à prendre en compte. tout comme le fichier .gitignore

yverhenne (commented):
Le fichier .gitignore n'est pas à prendre en compt

NicoDucou:
Hi @yverhenne 

A few remarks regarding this PR.
First there is a problem in the language ribon that lost its shadow (triangle shape). It looks like this normaly :
image

And in your branch the shadow is left on top alone : 
image

This PR would change the standard behaviour of Chamilo and we avoir doing changes that affect all the user by default.
Since the changes are only in css and tpl you could apply those changes on your plateform directly putting the tpl files in the overrides template folder and the css change in you personalized style sheet that you can import from the page  /main/admin/settings.php?category=Stylesheets

So this PR can not be included in Chamilo.

Best regards

PR #6499, opened 2025-07-22
Fix total courses progress calculation in mystudent.php

yverhenne:
Currently on the page mystudent.php, the calculation of the total progression in a session is based on the number of courses contained in this session with the formula: (total progressions/total number of courses in the session)x100. However, when a user is unenrolled from a course in the session using /main/session/session_course_user_list.php, the formula does not change, whereas the calculation should be made based on the number of courses the user is enrolled in this session with the formula: (total progressions/total number of courses where the user is enrolled in this session)x100.


NicoDucou:
Well spoted. All good just adapted the commit message.

PR #6531, opened 2025-07-29
Tracking:#Enable improved tracking section in main/mySpace/myStudents.php

yverhenne:
Option to enable in configuration.dist.php to add a progress summary section in the sessions, in mystudents.php


main/install/configuration.dist.php
```diff
@@ -1953,6 +1953,9 @@
 //Create an extra field for courses with identifier "theoretical_time"
 //$_configuration['display_theoretical_time'] = false;
 
+// Enable improved tracking section in main/mySpace/myStudents.php
+$_configuration['improve_tracking_in_mystudent_php'] = true;
```
NicoDucou: You have to put the line in comments (with // at the beginning) and the variable should be set to "false" by default.
The settings are always in comment otherwise in a new installation the parameter would be automatically set and we don't want that.
We also set the variable with the value that correspond to the default functioning of the platform. So in this case in a default configuration we do not show the synthesis part the setting should be set to false by default.

NicoDucou:
The language files main/lang/*/trad4all.inc.php should not be sent directly in your code. It has to be integrated through the translation platform https://translate.chamilo.org/
So you should remove those files from you PR.

yverhenne:
Variables de traduction.
main/lang/english/trad4all.inc.php :
$AverageProgressInSessions = "Average progress in sessions";
$TimeSpentInCoursesLastWeek = "Time spent in courses last week";
$ProgressionInSessions = "Session progress";
$synthesis = "Synthesis";
$WeeklyTimeSummary = "Weekly time spent summary";

main/lang/french/trad4all.inc.php :
$AverageProgressInSessions = "Progression moyenne dans les sessions";
$TimeSpentInCoursesLastWeek = "Temps passé dans les cours la semaine dernière";
$ProgressionInSessions = "Progression dans les sessions";
$synthesis = "Synthèse";
$WeeklyTimeSummary = "Synthèse des temps passés par semaine";

main/lang/spanish/trad4all.inc.php :
$AverageProgressInSessions = "Progreso medio en las sesiones";
$TimeSpentInCoursesLastWeek = "Tiempo dedicado en los cursos la semana pasada";
$ProgressionInSessions = "Progreso en las sesiones";
$synthesis = "Síntesis";
$WeeklyTimeSummary = "Resumen de tiempo dedicado por semana";

Issue #4283, opened 2022-05-16
Default "System" user


ywarnier:
In Chamilo 1.11 (and earlier), there is a recurrent issue when a user with privileges (admin being the easiest example) is deleted, as he might have created other users, which then have the "creator_id" set to the deleted user, and inconsistencies arise in some reports after that.
There is also an issue with sending notifications from the portal, as these are usually (initially) associated with the account of the admin who installed the portal.

By creating a "System" account that cannot be deleted and which nobody can "login" into, we could make sure that resources attached to an admin are automatically associated with this "System" account when the admin user is deleted, and we could handle sending notifications in a more neutral way.

Adding this type of user requires an additional field ("can_login") in the user table, and a new user type (called "status" in the user table) "system" (i.e. 99) in the list of user types. This should probably *not* be inserted in the "role" table, to avoid tampering.

This user should also be created automatically during installation and upgrades from 1.11.x.

(BT#19859 is such a case of something going wrong because we lack this mechanism)

christianbeeznest:
Hi @ywarnier ,

Since DB schema changes aren’t allowed in beta, should we implement the “System” user using logic only (e.g. deny login if username = 'system', prevent deletion by ID), without adding can_login or a new status field?

Let me know if that’s acceptable or if you'd prefer to postpone this.

ywarnier:
Hi @christianbeeznest 
Actually this is much easier now than initially planned: please simply use the "fallback_user" for that. That should solve the issue similarly to what was described above : the fallback_user is already a user without role and a status of 99, planned right for this.

christianbeeznest:
It's implemented with this PR https://github.com/chamilo/chamilo-lms/pull/6535

Thanks for confirmation.

christianbeeznest:
Clarifying scope: we’ll keep fallback_user strictly for internal ownership/creator reassignment and as the neutral sender inside the DB.
SMTP headers will always use platform settings (From + default Reply-To = platform), with an optional mail.reply_to_policy=sender for explicit opt-in. No UI to edit fallback_user e-mail.

ywarnier:
Note : the user.active column is set to "-2" and user.status is set to 99 to differenciate it from other users. user.username = 'fallback_user'. Usually, it gets user.id=3 when installing a new portal as the installer admin gets id=1 and "anon" gets id=2.

ywarnier:
@christianbeeznest This implementation is incomplete:
- login as admin
[...]

PR #6535, opened 2025-07-30
Internal: Use fallback_user as email sender and set Reply-To to actual sender - refs #4283

christianbeeznest:


ywarnier:
Looks good to me. I'm not sure what the e-mail of the fallback user is, though (where is it configured?)

ywarnier:
Although the code looks good, this doesn't seem to be what I would be expecting.
- There is no place/process where we suggest to update the fallback user's e-mail address.
- Sending with the sender's "reply-to" should not be the behavior by default (we should use the platform's address as reply-to, otherwise we encourage off-platform conversations, which we generally want to avoid)

So I'm not merging this PR for now. We need to discuss this more.

christianbeeznest:
Closing in favor of a revised approach aligned with #4283.
We should not use fallback_user in SMTP headers. Instead:

From/Reply-To come from platform settings (default Reply-To = platform).

fallback_user is used only for internal DB associations (ownership/creator reassignment, neutral sender inside Chamilo).

Issue #4763, opened 2023-06-23
[C2] Learning path tool home page (LP list) new interface


NicoDucou:
The home page of the Learning path tool is in legacy like this : 
![image](https://github.com/chamilo/chamilo-lms/assets/6579268/1126ad8e-bff3-42a0-b22f-3666e16c1f83)

And it should be integrated with vue to look like this : 
![image](https://github.com/chamilo/chamilo-lms/assets/6579268/92c85ed7-6c45-4d8d-b516-4cb5db3d2875)

Most of the information is already present in the current situation  except from the image and the prerequisities.

To have LP you need to create them here : main/lp/lp_controller.php?cid=2&sid=0&gid=0&gradebook=0&origin=&action=add_lp_category

And to have LP you have to create them here : main/lp/lp_controller.php?cid=2&sid=0&gid=0&gradebook=0&origin=&action=add_lp

christianbeeznest:
It is implemented a version about it in this PR https://github.com/chamilo/chamilo-lms/pull/6589 , it requires more testing to be sure all functionality are Ok

Thanks for confirmation.

ywarnier:
This has been approved and integrated. Improvements have been requested through #6609

PR #6589, opened 2025-08-13
Learnpath: Implement tool home page (LP list), new interface - refs #4763

christianbeeznest:


ywarnier:
I'm merging this because it works, but there are a number of things that need to be improved. I'm documenting those in #4763

Issue #5578, opened 2024-06-07
[C2] Enable logo changes with custom styles


ywarnier:
Still missing in C2.
The logo should go in the custom-CSS directory. More details in private issue.

Refs BT#21621

ywarnier:
For now (with https://github.com/chamilo/chamilo-lms/pull/5614) the logo has to be added manually in var/themes/[your-theme]/images/header-logo.png in 180x40 pixels, but it works. We're still in alpha, so that's cool for now.

ywarnier:
This issue evolves a little bit: it should be possible to upload 4 files (all optional):
- platform logo in SVG (var/themes/[the-theme]/images/header-logo.svg)
- platform logo in PNG (var/themes/[the-theme]/images/header-logo.png)
- email logo in SVG (var/themes/[the-theme]/images/email-logo.svg)
- email logo in PNG (var/themes/[the-theme]/images/email-logo.png)

The SVG file is used in priority, if it is present.

The header logo in PNG must be max 190px wide and 60px high.

The recommended email logo width is 540px as per Chamilo 1.11 (`$_configuration['email_logo']`) (and 170px height if we respect the proportions of header-logo.png). In Chamilo 2, as soon as there is a file in there, it is served as e-mail logo.

christianbeeznest:
- 4 optional logos per theme in var/themes/<slug>/images/: header-logo.svg|png, email-logo.svg|png (SVG takes priority).

- Rules & endpoints: header PNG ≤ 190×60; email PNG ~540px (advised). POST /themes/{slug}/logos, DELETE /themes/{slug}/logos/{type}, GET /themes/{name}/{path}?strict=1&t=… (creates /images, sanitizes SVG, cache-bust).

- UI: AdminBranding shows effective + per-file previews and Remove, with fallback SVG→PNG→default; PlatformLogo uses the same fallback and default logo.

All included in PR: https://github.com/chamilo/chamilo-lms/pull/6602

Thanks for confirmation.

ywarnier:
This works but it should be inside the "colors" page. Otherwise you're not sure which CSS theme you're modifying with logos.
In my case I created a new theme, saved it, then went to "Branding", deleted the default logos, uploaded mine, and then my logo was in place of the default style in the var/themes/chamilo/images/ folder. It should have been saved in var/themes/[mytheme]/images/

An alternative (to avoid rewriting files) would be to show clearly what theme the logos will be applied to, inside the branding page.

[...]

PR #6602, opened 2025-08-20
Admin: Enable logo changes with custom styles - refs #5578

christianbeeznest:


ywarnier:
This works but it should be inside the "colors" page. Otherwise you're not sure which CSS theme you're modifying with logos.
In my case I created a new theme, saved it, then went to "Branding", deleted the default logos, uploaded mine, and then my logo was in place of the default style in the var/themes/chamilo/images/ folder. It should have been saved in var/themes/[mytheme]/images/

Issue #6610, opened 2025-08-21
[C2] Top menu bar should check existing pages


ywarnier:
The current top bar menu when logged out, in Chamilo 2, shows 6 entries:

<img width="588" height="92" alt="Image" src="https://github.com/user-attachments/assets/efef526c-a488-4371-94f1-602c456bc3b6" />

These entries always show at the moment, but they should depend on existing pages (from the `page` table) combined with page categories.

This is a list of what has to be checked before the entries are shown in TopBarNotLoggedIn.vue:

| Entry | Process |
| --- | --- |
| Home | SELECT count(id) FROM page JOIN page_category on page.category_id = page_category.id WHERE page_category.title = 'index' AND page.enabled=1 AND page.locale='{CURRENT_LOCALE}' AND access_url_id = {CURRENT_URL} |
| FAQ | SELECT count(id) FROM page JOIN page_category on page.category_id = page_category.id WHERE page_category.title = 'faq' AND page.enabled=1 AND page.locale='{CURRENT_LOCALE}' AND access_url_id = {CURRENT_URL} |
| Registration | SELECT count(*) FROM settings WHERE variable = 'allow_registration' AND selected_value = 'true' AND access_url_id = {CURRENT_URL OR GLOBAL IF NOT CURRENT} - use the designed methods for that |
| Demo | SELECT count(id) FROM page JOIN page_category on page.category_id = page_category.id WHERE page_category.title = 'demo' AND page.enabled=1 AND page.locale='{CURRENT_LOCALE}' AND access_url_id = {CURRENT_URL} |
| Contact | Check if the contact page is on for this URL and this locale (not sure of the process here) |
| Language | SELECT count(*) FROM language WHERE available = 1 (only show if > 1)|

This requires a provider or some construct like that.

christianbeeznest:
It is implemented with this PR https://github.com/chamilo/chamilo-lms/pull/6625

Thanks for confirmation.

ywarnier:
This is working as expected (I didn't test the language selector's visibility though - trusting that one works as well as the others). Closing.

PR #6625, opened 2025-08-25
Internal: Topbar: derive visibility from pages by category/URL/locale - refs #6610

christianbeeznest:


ywarnier:
This is working well. Merging.

PR #6672, opened 2025-09-02
Add sorting by date and search by user in pending.php

yverhenne:
Add sorting by date to pending.php
Add search by user to pending.php


NicoDucou:
Thank you @yverhenne for your contribution.

Issue #6686, opened 2025-09-04
[C2] Remove "enabled" settings from plugins


ywarnier:
Some of the plugins that have been transfered from C1 have an "tool_enable" setting in their installation process.
This "enable" is now replaced by access_url_rel_plugin.active, so keeping the "tool_enable" setting just adds confusion.
Make sure the setting itself is not used anymore (or change it to use access_url_rel_plugin.active) and then remove it from the plugin.

This is an example of the changes needed https://github.com/chamilo/chamilo-lms/commit/540590e3ed081c4d9efe08e744e082c0cad79813

No migration is necessary as plugins migration is not supported from C1 to C2.0.

So tasks:
- [ ] Scan all plugin installers to see if there is a "tool_enable" setting created
- [ ] For each of these, scan the Chamilo code to find if it is used. If so, replace by the following block
- [ ] Update the plugin to not create the "tool_enable" setting (maybe it has a slightly different name - there was no convention for that in 1.11)

```
    $plugin = $pluginRepo->findOneByTitle('zoom');
    $pluginConfiguration = $plugin?->getConfigurationsByAccessUrl(Container::getAccessUrlUtil()->getCurrent());
    $isInstalled = $plugin && $plugin->isInstalled();
    $isEnabled = $plugin && $pluginConfiguration && $pluginConfiguration->isActive();

    if ($isInstalled && $isEnabled) {
```

christianbeeznest:
It is checked with this PR https://github.com/chamilo/chamilo-lms/pull/6703 

Thanks for confirmation.

PR #6703, opened 2025-09-05
Plugin: Remove "enabled" settings from plugins - refs #6686

christianbeeznest:


ywarnier:
Good job. The plugins still include a mention of tool_enable in their __construct() method, which shows the "Enable" option despite not having the database record, but I'm fixing this in a coming change.

Issue #4540, opened 2023-01-15
Edit default listing number (always 20)


IkkeMus:
Every time I visit my Excercises, the Results and feedback of Excercises, the User list and the Glossary I get a default list of 20. 
Most classes and groups are 25-30 students or more.
So I have to change this to 50 or All. It is just an inconvenience, but I do this approximately a 100 times each day, if not more.

Would it be possible to edit the default setting in the admin page (platform administration) or course settings to a higher number/All? (even 50 or 100 would help a lot) 

It will reduce the stress on the system resources as the teacher and student will not have to load the same page twice.

(btw as a newbie I absolutely love Chamilo, specially the excercises module, looking forward to version 1.11.18.)

ywarnier:
Hi @IkkeMus 
We're not going to be able to deliver this in 1.11.18 (about to be released) but we will consider it for future versions. Sorry to hear it is so frustrating, we definitely didn't have that one on our radar as something important to anyone. Noted.

NicoDucou:
Hi @IkkeMus 
There are already some configuration option that can be activated  in app/config/configuration.php for some tables in Chamilo : 

// Show pagination if question list is bigger than "x" value, if 0 pagination will not appear.
// Option only when building an exercise as a teacher
// $_configuration['show_question_pagination'] = 100;

// Number of questions to show in every page
// Option only when building an exercise as a teacher
// $_configuration['question_pagination_length'] = 20;

// Changes the row list when using jqgrid/sortable tables
//$_configuration['table_row_list'] = ['options' => [50, 100, 200, 500]];

// Default selected row in jqgrid/sortable tables
//$_configuration['table_default_row'] = 50;

@IkkeMus I let you check those options because some of the table that you are talking about correspond to those settings, maybe not all but you will already be able to have a solution for some of them.

ywarnier:
Update to this issue : check the options listed by Nicolas are available and work as intended.

christianbeeznest:
Hi @ywarnier,

I’ve confirmed that the new settings are already applied in the code:

Platform Administration → Settings

Table row list (table_row_list) defines the available options (e.g. 50, 100, 200, All)

Default table rows (table_default_row) sets the default value
These are now read by all grid_js tables instead of using the hard‑coded “20.”

Course → Exercises → Settings

Show question pagination (show_question_pagination)

Question pagination length (question_pagination_length)
The exercise builder’s question lists likewise pull their limits from these settings.

[...]

PR #6747, opened 2025-09-13
Internal: Respect BaseTable row settings; default 0 to 20; fix pagination

christianbeeznest:


christianbeeznest:
refs #4540

ywarnier:
Good for me.
Note: when changing it as admin, the change might not appear immediately (and require a logout/login sequence).

Issue #6761, opened 2025-09-16
[C2] Restore global chat


ywarnier:
The global chat feature from 1.11 has been disabled but still appears in the settings.

This issue is for 3 different (related) things:
- Add a migration and change datafixtures to ensure that the allow_global_chat is false by default and hide_chat_video is true by default (this will make the settings more consistant with normal usage)
- Restore the global chat feature using Vue (there might exist components doing most of the work already). This uses the "chat" and "chat_video" tables
- Look for Open Source libraries that would allow us to re-implement video chat. Initially, we were using SimpleWebRTC, but its license changed to an incompatible license for Chamilo, so we had to abandon it. Warning note: WebRTC only works with HTTPS, so don't try to test it on HTTP

Only the first item is necessary for Beta. The rest can come afterwards, but it would be interesting to include any new dependency on a WebRTC library (if any matches our needs) in the Beta as well.

christianbeeznest:
We have a version of global chat with this PR https://github.com/chamilo/chamilo-lms/pull/6805

Thanks for confirmation.

ywarnier:
Tested and approved

PR #6805, opened 2025-09-25
User: Restore global chat - refs #6761

christianbeeznest:


ywarnier:
Very good on your first attempt! Merging.

Capture d’écran du 2025-09-26 15-24-58

PR #6817, opened 2025-09-26
Install: Fix language selection flow, i18n sync, and labels

christianbeeznest:


ywarnier:
I'm merging it *but* I still see 2 issues:

- when my browser is configured to support the following languages, in this order: `fr, fr_FR, en_US`, when I'm loading Step1 with no cookies yet, I see this:
install-home
Obviously, there is some mix-up between the preferred browser language and the available languages available in Chamilo. In any case, it shouldn't show `fr` in the select box...

- the Portuguese language (`pt_BR`) does not show in the drop-down list, despite it being in Step1.vue. I'm not sure why that would be.

PR #6840, opened 2025-09-30
Internal: Update extrafield flags for 'my_terms','terms_ville','terms_villedustage' not included in migration

christianbeeznest:


ywarnier:
To be  executed on an installed system with: `php bin/console doctrine:fixtures:load --group=extrafield-update --append`

PR #6859, opened 2025-10-03
Migrate LTI tables from plugin to core

AngelFQC:


ywarnier:
When I execute the 3 migrations (Version2025100211000[0,1,2] on my system with Chamilo master installed in dev mode *before* those new migrations, I get errors only for [...0]):
```
[notice] Executing Chamilo\CoreBundle\Migrations\Schema\V200\Version20251002110001 up
^[[A[error] Migration Chamilo\CoreBundle\Migrations\Schema\V200\Version20251002110001 failed during Execution. Error: "An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1553 Cannot drop index 'IDX_5C76B75D8F7B22CC': needed in a foreign key constraint"                                                                                                                                                                  

In ExceptionConverter.php line 118:
                                                                                                                                                                    
  An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1553 Cannot drop index 'IDX_5C76B75D8F7B22CC': needed in a foreign key constraint  
                                                                                                                                                                    

In Exception.php line 24:
                                                                                                                     
  SQLSTATE[HY000]: General error: 1553 Cannot drop index 'IDX_5C76B75D8F7B22CC': needed in a foreign key constraint  
                                                                                                                     

In Connection.php line 71:
                                                                                                                     
  SQLSTATE[HY000]: General error: 1553 Cannot drop index 'IDX_5C76B75D8F7B22CC': needed in a foreign key constraint  
```

So it seems IDX_5C76B75D8F7B22CC is still used somewhere when you're trying to drop it.
I would imagine this is not super problematic, but then I'm not sure if it executes the rest of the migration. At least in my case it didn't drop `lti_lineitem.tool_id`.

I get this error on several fields: If I try to fix the first one, I get a message reminding us that the 'public_key' index already exists, etc. Something seems wrong.

There are also a number of queries like `"ALTER TABLE DROP COLUMN [...]"` (!!) where the name of the table is not stated (my IDE flags them as erroneous queries, with reason).

There's no "down()" method in this migration either, so it's a bit tricky to turn back once you ran the script and *some* of it worked.

On line ~262 of Version20251002110001.php, I find the use of $token->hasColumn('public_key') while the object it is executed on should really be $platform.

Issue #6814, opened 2025-09-26
[C2] Delete course from course_list.php when course is in sequence doesn't work


ywarnier:
### How to reproduce

- Create 2 courses
- Create a resource sequence with those 2 courses in it
- Go to main/admin/course_list.php
- Try do delete any of those 2 courses
- The page refreshes but the course is not deleted and no error message appears
- Go to one of the courses' home
- Click "Course maintenance" (cog)
- Click "Delete this course" and add the course code and confirm
- An error message appears telling you that the course is included in a sequence and that you need to remove it from there first. This is the desired behaviour.

### What to change

Update course_list.php to show the same error message when the course is in a sequence.

ywarnier:
<img width="1274" height="363" alt="Image" src="https://github.com/user-attachments/assets/f19d64f3-1a82-4f54-b984-25f4035ee572" />

Works as requested now. Thanks!

PR #6868, opened 2025-10-07
Fix: Flashes messages were not displayed for some actions - ref #6814

Coussecousse:
Hi!

**Issue:**
[https://github.com/chamilo/chamilo-lms/issues/6814](https://github.com/chamilo/chamilo-lms/issues/6814)

The flash messages were not displayed on the `course_list` page. It was the case when you deleted a course or when you click on the category button.

**Problem:**
When using `api_location(api_get_self());`, the flash messages were somehow cleared.
I tried using the PHP `header()` function or attempted to manually write the session with `session_write_close()`, but nothing worked.
The only way to make it work was to remove this line, but doing so prevented the page from redirecting — leaving the parameters visible in the URL, which I’m not very comfortable with (especially the `sec_token`).

**Resolution:**
I changed the actions to use a form with the POST method, and now everything works correctly.
I also removed the `'text-warning'` class because it didn’t work before, and now that it does apply, it looked a bit odd :

Capture d’écran du 2025-10-07 11-42-07

**Proof:**
[Capture vidéo du 2025-10-07 11-46-38.webm](https://github.com/user-attachments/assets/33f5e234-318b-46d5-a07a-612ebbfd8c9f)


AngelFQC:
Flash messages in public/main/admin/course_list.php fixed here 4c0750e941da3bf286941bdfb791128c73aa8e92

public/main/course_info/delete_course.php is not used with the new interface

Issue #6738, opened 2025-09-11
[C2] Health check box on admin page


ywarnier:
Some issues on a new Chamilo portal can be difficult to diagnose if they are stemming from a bad configuration.

I think we should add a block on the admin page called "Health check" (as many other software do) and show a few indicators there, starting with whether all settings *required* for e-mail sending are effectively configured.

The block would just show a list of items (&lt;ul&gt; list) with a green check or a red cross if not properly configured, and a link to configure them properly.

For example, 
- [x] if the mail DSN or the origin sender e-mail are not defined, it should say so and link to /admin/settings/mail (text: "E-mail settings are OK")
- [x] if there are more than one active URL and the global admin is not in all of them, it should point to /main/admin/access_url_edit_users_to_url.php (text: "Admin has access to all URLs")

ywarnier:
Tested and approved

PR #6879, opened 2025-10-08
Feat: Add health check box on admin page- ref #6738

Coussecousse:
Hi !

I just added a health check box just as you requested (I think) in https://github.com/chamilo/chamilo-lms/issues/6738 :smile: 

**Check if :** 
- The mail DSN or the origin sender e-mail are not defined
- If there are more than one active url, the global admin should be in it. 

**Screens :** 
- Nothing is good :disappointed:  :
Capture d’écran du 2025-10-08 14-45-38

- URL is active and Admin is not register in it (: 
Capture d’écran du 2025-10-08 14-21-04

- Email is ok : 
Capture d’écran du 2025-10-08 14-20-51

- All is good :smile: : 
Capture d’écran du 2025-10-08 14-20-22

________________________________
:warning: I don't know why, but `mailer_from_email` seems to be mixed up... 
**Proofs :** 
Capture d’écran du 2025-10-08 14-47-52
Capture d’écran du 2025-10-08 14-41-43

I noticed it when I was doing my test. I will investigate


ywarnier:
For `mailer_from_email`, we knew and updated it

public/main/inc/lib/api.lib.php
```diff
@@ -5259,6 +5259,43 @@ function api_get_access_url_from_user($user_id)
...
+        return true;
+    }
+}
+
+/**
+ * Gets all the access urls in the database.
+ *
+ * @return array
+ */
+function api_get_active_urls()
```
ywarnier, line 5286: Instead of adding api_get_active_urls() here, you should define a similar method in src/CoreBundle/Helpers/AccessUrlHelpers.php
Then you could call it through:
```
$accessUrlHelper = Container::$container->get(AccessUrlHelper::class);
$currentUrl = $accessUrlHelper->getCurrent();
```

Beware that these will return AccessUrl objects, which have to be called a bit differently than you code here.

public/main/admin/access_urls.php
```diff
@@ -79,13 +79,14 @@
 $parameters['sec_token'] = Security::get_token();
 
 // Checking if the admin is registered in all sites
-$url_string = '';
-foreach ($url_list as $u) {
-    if (!in_array($u->getId(), $my_user_url_list)) {
-        $url_string .= $u->getUrl() . '<br />';
+if (!api_is_admin_in_all_active_urls()) {
```
ywarnier, line 82: The check should be on whether *any* user with GLOBAL_ADMIN privileges is registered in each URL. The current user account is not really important, but at least one global admin should be registered in each URL.

So the method `api_is_admin_in_all_active_urls()` should rather be called `api_each_url_has_admin()` or something like that. And ideally, this should be defined as a method of the AccessUrlHelper (see below).

ywarnier (approved):
Some improvements to be made later on, but overall OK. Thanks !

Issue #6344, opened 2025-06-10
In Course>Users>Usergroups, add a action button for usergroups to have the preview list members


chamilo-uga:
Hi,
the goal is to allow teachers to quickly have a full users list page of members of a class registered to a course.
It is a very simple and basic overview page with <ul><li> list of users and a search bar.
It's voluntary on our part to have a sober large vertically scrolled page with no jqGrid and just a simple search bar on top.
There are 2 lists
1. Students in class registered in course
2. Students in class unregistered in course(*)

(*) a teacher can unregister from a course a student in a registered class in course.

We have information 
- student firstname + lastname and email
because users dont often know the username of a user in platform.
It is easier for them to use email to check for homonymous

![Image](https://github.com/user-attachments/assets/055442a0-94a8-493a-86c8-e516caafc74c)

ywarnier:
Merged into f7ec08ae7a390c10e1fde3a7a6f7f7d93686533f with a number of additional fixes.
Sorry about the attribution, I messed up and it appears as mine.
Thanks @chamilo-uga

PR #6880, opened 2025-10-08
#6344 - In Course>Users>Usergroups, add a action button for usergroups to have the preview list members

chamilo-uga:
See https://github.com/chamilo/chamilo-lms/issues/6344
---
the goal is to allow teachers to quickly have a full users list page of members of a class registered to a course.
It is a very simple and basic overview page with

    list of users and a search bar.
    It's voluntary on our part to have a sober large vertically scrolled page with no jqGrid and just a simple search bar on top.
    There are 2 lists

        Students in class registered in course
        Students in class unregistered in course(*)

    (*) a teacher can unregister from a course a student in a registered class in course.

    We have information
        student firstname + lastname and email
        because users dont often know the username of a user in platform.
        It is easier for them to use email to check for homonymous

### DIFF
--- a/public/main/inc/lib/usergroup.lib.php
+++ b/public/main/inc/lib/usergroup.lib.php
@@ -12,7 +12,6 @@ use Chamilo\CoreBundle\Service\StandardizationService;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Polyfill\Intl\Normalizer\Normalizer;
 
-
 /**
  * Class UserGroup.
  *


public/main/inc/lib/usergroup.lib.php
```diff
@@ -1209,6 +1211,36 @@ public function unsubscribe_courses_from_usergroup($usergroup_id, $delete_items,
         }
     }
 
+    public function unsubscribe_only_courses_from_usergroup($usergroup_id, $delete_items, $sessionId = 0)
```
ywarnier, line 1214: PHPDoc block needed to document the method

ywarnier: Method should provide a return type

public/main/inc/lib/usergroup.lib.php
```diff
@@ -1209,6 +1211,36 @@ public function unsubscribe_courses_from_usergroup($usergroup_id, $delete_items,
...
+    {
+        $sessionId = (int) $sessionId;
+        // Deleting items.
+        if (!empty($delete_items)) {
+            $user_list = $this->get_users_by_usergroup($usergroup_id);
+
+            $groupId = isset($_GET['id']) ? (int) $_GET['id'] : 0;
+            foreach ($delete_items as $course_id) {
+                $course_info = api_get_course_info_by_id($course_id);
+                if ($course_info) {
```
ywarnier, line 1224: This is inefficient. `api_get_course_info_by_id()` is a heavy function. Please just replace by a validation that $course_id is an `int`, and that should be enough (worst case: the course_id is not present in usergroup_rel_course and the delete fails but does not break the process).

public/main/inc/lib/usergroup.lib.php
```diff
@@ -1209,6 +1211,36 @@ public function unsubscribe_courses_from_usergroup($usergroup_id, $delete_items,
...
+                        [
+                            'usergroup_id = ? AND course_id = ?' => [
+                                $usergroup_id,
+                                $course_id,
+                            ],
+                        ]
+                    );
+                }
+                if (0 != $sessionId && 0 != $groupId) {
+                    $this->subscribe_sessions_to_usergroup($groupId, [0]);
```
ywarnier, line 1236: You cannot add a "subscribe*()" call inside a method that is called "unsubscribe_only_courses_from_usergroup()". What is the goal here ?

public/main/inc/lib/usergroup.lib.php
```diff
@@ -1209,6 +1211,36 @@ public function unsubscribe_courses_from_usergroup($usergroup_id, $delete_items,
...
+                } else {
+                    $s = $sessionId;
```
ywarnier, line 1238: This `$s` is never used. It should either be removed or used and documented.

src/CoreBundle/Service/StandardizationService.php
```diff
@@ -0,0 +1,72 @@
+<?php
```
ywarnier, line 1: This should be a Helper

src/CoreBundle/Controller/ClassController.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
ywarnier, line 1: This code should go into a new file called UsergroupController.php rather than ClassController.php (we don't use the term "Class" in the code anymore).

src/CoreBundle/Controller/UserController.php
```diff
@@ -8,21 +8,48 @@
...
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Routing\Attribute\Route;
 use Symfony\Component\Security\Core\User\UserInterface;
+use UserGroupModel;
 
 /**
  * @author Julio Montoya <gugli100@gmail.com>
  */
-#[Route('/user')]
```
ywarnier, line 19: Removing `/user` is not a super nice idea. Better leave it and use `/user/overview` as path for `overview()`.
Paths with "main" are *only* kept until all the legacy code is migrated to Vue

src/CoreBundle/Resources/views/Class/overview.html.twig
```diff
@@ -0,0 +1,98 @@
+{% extends "@ChamiloCore/Layout/base-layout.html.twig" %}
```
ywarnier, line 1: Please move to src/CoreBundle/Resources/views/Usergroup/overview.html.twig

public/main/inc/lib/usergroup.lib.php
chamilo-uga: $course_info['code'] is used in function CourseManager::unsubscribe_user we use

src/CoreBundle/Controller/ClassController.php
```diff
@@ -0,0 +1,30 @@
...
+namespace Chamilo\CoreBundle\Controller;
+
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Attribute\Route;
+use UserGroupModel;
+
+class ClassController extends AbstractController
+{
+    #[Route('/main/admin/classes/{id}/overview', name: 'class_overview')]
```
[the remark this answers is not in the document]
AngelFQC: And the controller file should located in `src/CoreBundle/Controller/Admin`

Issue #6883, opened 2025-10-08
[C2] organisationemail and other extra fields duplicated or badly titled


ywarnier:
For some reason, `ExtraFieldFixtures.php` contains the `organisationemail` twice for user fields: once as `organisationemail` directly, and another one as `AzureAuthenticatorHelper::EXTRA_FIELD_ORGANISATION_EMAIL`.

The second one is clearly needed for Azure authentication. The first one... not clear, so I would remove the first one, but this would require changing ExtraFieldFixtures.php *and* adding a migration to remove the duplicate.

This is also the case for `azure_id` (remove the first one).

Also, the *display_text* for:
- `LinkedInUrl` should say `LinkedIn profile URL`
- `VideoUrl` should say `Video URL`
- `tags` should say `Tags`
- `Legal` should say `Legal agreement accepted`
- `GDPRTitle` should say `GDPR`
- `GDPRTextArea` should say `GDPR terms to be shown to users`

Also, it would be necessary to have some kind of validation for extra_field.variable + extra_field.item_type to be unique, otherwise we end up with several fields with the same name.

But for alpha 5, we need at least the fix to remove the duplicate `organisationemail`.

christianbeeznest:
It is checked with this PR https://github.com/chamilo/chamilo-lms/pull/6884

Thanks for confirmation.

ywarnier:
Works for me.

PR #6884, opened 2025-10-08
Internal: ExtraField: deduplicate organisationemail/azure_id (USER), fix display_text, and add unique index on (variable, item_type) - refs #6883

christianbeeznest:


ywarnier:
Thanks!

PR #6897, opened 2025-10-10
Fix: /sessions give error 500

Coussecousse:
Hi ! 

While running tests for the chart documents, I noticed that the page `/sessions` throws a 500 error.

**Screens** : 
Capture d’écran du 2025-10-10 15-56-43
Capture d’écran du 2025-10-10 15-56-32

I applied a small fix for it.

**Screen after fix :** 
Capture d’écran du 2025-10-10 15-57-05


ywarnier:
Just a heads up: the whole team is currently busy on other aspects of Chamilo, so we're not looking much at PRs this week, but will retake that next week. Sorry for the delay reviewing.

christianbeeznest:
OK to merge as-is — paginationEnabled:false is already configured in API Platform and matches the provider’s array return.

Issue #6837, opened 2025-09-30
[C2] Quiz: reimplement export all result functionnality from 1.11.x


NicoDucou:
In Chamilo 2 on the administration page in the "Sessions management" section there is a "Export all results from an exercise" entry that points to /main/admin/export_exercise_results.php 
The file public/main/admin/export_exercise_results.php does not exist in Chamilo 2. It is present in Chamilo 1.11.x
This issue is the import this file in Chamilo 2 and adapt it if necessary to make the functionnality work in Chamilo 2.

PR #6918, opened 2025-10-17
Quiz: reimplement export all result functionnality from 1.11.x - refs #6837

Coussecousse:
Hi!

I reimplemented the old file `export_exercise_results.php` for Chamilo v2.
I'm using repositories to find exercises, courses, etc. Hope that's fine.
In the old version, the queries were not dynamic, so I stuck with the same logic, except for the fact that I'm using Select2 instead of Selectpicker (I didn’t manage to use it properly — it seems it's not implemented yet in Chamilo v2).

You also can't select a course if you haven’t selected a session first, and you need a course selected to choose an exercise.

I’m not sure if I’ve tested every possible case, so feel free to do so and let me know if you need any readjustments.

**Issue: https://github.com/chamilo/chamilo-lms/issues/6837**

**Screens:**
Capture d’écran du 2025-10-17 09-47-56

Capture d’écran du 2025-10-16 16-38-34 Capture d’écran du 2025-10-16 16-38-27

**:warning: PDF result :**
Capture d’écran du 2025-10-16 15-22-42

(I figure it should be fixed in another PR.)

### DIFF
--- a/src/CoreBundle/Framework/Container.php
+++ b/src/CoreBundle/Framework/Container.php
@@ -631,11 +631,6 @@ class Container
         return self::$container->get('form.factory');
     }
 
-    public static function getCQuizRepository(): CQuizRepository
-    {
-        return self::$container->get(CQuizRepository::class);
-    }
-
     public static function addFlash(string $message, string $type = 'success'): void
     {
         $type = match ($type) {


public/main/admin/export_exercise_results.php
```diff
@@ -0,0 +1,237 @@
...
+// Get request parameters
+$sessionId = isset($_REQUEST['session_id']) ? (int) $_REQUEST['session_id'] : null;
+$courseId = isset($_GET['selected_course']) ? (int) $_GET['selected_course'] : null;
+$exerciseId = isset($_REQUEST['exerciseId']) ? (int) $_REQUEST['exerciseId'] : null;
+
+// Get repositories
+$user = api_get_user_entity(api_get_user_id());
+$sessionRepository = Container::getSessionRepository();
+$courseRepository = Container::getCourseRepository();
+$cQuizRepository = Container::getCQuizRepository();
```
christianbeeznest: Container already exposes getQuizRepository(): CQuizRepository. Adding getCQuizRepository() introduces redundant naming for the same service.

src/CoreBundle/Framework/Container.php
```diff
@@ -631,6 +631,11 @@ public static function getFormFactory(): FormFactory
         return self::$container->get('form.factory');
     }
 
+    public static function getCQuizRepository(): CQuizRepository
```
christianbeeznest: Remove getCQuizRepository() and use getQuizRepository() everywhere for consistency

PR #6959, opened 2025-10-28
Attendance: breadcrumb create fix — sheet is link, form is plain (respect meta.breadcrumbLink=false)

christianbeeznest:


ywarnier:
This solution is interesting but it would imply adding a new attribute to any "last step" appearing in the Breadcrumb and I just can't see that being practical for now. We might have to come back to this later on if we need more control over the breadcrumb, but for now, https://github.com/chamilo/chamilo-lms/commit/e7e1d53f6e41a4ca43cbc5575d00a95159115f18 will do just fine. Thanks!

Issue #5627, opened 2024-07-03
[C2] In the forum thread, answering shows thread with top bar


ywarnier:
When answering any thread in any forum, the "Thread" area below the message section shows the other messages of this thread... with the Chamilo top bar covering about 2 messages (impossible to read them).

ywarnier:
![Capture d’écran du 2024-07-03 21-34-12](https://github.com/chamilo/chamilo-lms/assets/609839/e24d5799-1d78-4471-9791-0e196e3ab196)

PR #7004, opened 2025-11-06
Thread: fix top bar overlapping thread messages refs #5627

Ilyassbennanii:


ywarnier:
This PR was OK-ish (the CSS should have been in a central file) but there was some development on the side by @christianbeeznest (mostly in 0e2a4203e8299ad99b202540d9735f0866fd1afa) that actually solved this by removing the iframe and making the thread view in a div.

This is looking great now
Capture d’écran du 2025-11-14 10-02-11

(and the thread view link sends us to the normal thread view)
Capture d’écran du 2025-11-14 10-02-40

Issue #5744, opened 2024-08-21
[C2] when editing a LP when creating a new document, clicking 2 times on save button create 2 items


NicoDucou:
On the 2024/08/21 in chamilo Master when editing a LP there is no token to avoid the creation of 2 items when clicking on save.
To reproduce :
* create a LP
* click on "Create a document"
* fill in the form
* click 2 or more times on the Save button
You have 2 or more item duplicated included in the LP.

There should be a token to avoid sending many times the same form.

PR #7005, opened 2025-11-06
LP: fix duplicate LP items on multi-click save refs #5744

Ilyassbennanii:


ywarnier:
Tested and approved. Thanks!

PR #7042, opened 2025-11-12
Extended question pool plugin added

nosolored:
**Extended Question Pool plugin**
This plugin expands the functionality of the question pool.

_Installation instructions_
- Enable the plugin in the Chamilo administration panel.
- Configure the desired options

_Access the tool_
- From plugins, to configure the options
- Features included:
   - Display the total number of questions in the exercise
   - Extra fields in questions to expand the classification
   - Detect previously used questions
   - Assign scores to questions in bulk
   - Count correct and incorrect answers per question
   - Sort questions by name, category, difficulty, number of occurrences, correct answers, and incorrect answers

### DIFF
--- a/main/exercise/question_pool_extended.php
+++ b/main/exercise/question_pool_extended.php
@@ -5,15 +5,17 @@
 use Chamilo\CoreBundle\Entity\ExtraField as ExtraFieldEntity;
 use ChamiloSession as Session;
 use Knp\Component\Pager\Paginator;
+use Twig\Cache\NullCache;
 
 /**
- * Question Pool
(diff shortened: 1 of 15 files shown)


ywarnier:
This should be integrated as an extension to the question pool rather than as a plugin (the fact that main/exercise/question_pool_extended.php is so long and is not in the plugin folder confirms that relatively well).

The QUERY_STRING direct user in the header() call is a probable vulnerability.

The installation process inserts a series of extra fields which could be inserted by hand by anyone wanting to use those fields on questions.

Issue #5114, opened 2024-01-24
 Add setting to limit number of students subscribed to a course


ywarnier:
Some portals open to the world are being abused by users registering more than a hundred users to courses, while the portal is only meant for academical use (which typically spans 15 to 30 students).

Add a configuration setting:
```
// Set a maximum number of users per course, globally. Set to 0 to remove the limit.
$_configuration['hosting_limit_users_per_course'] = 0;
```
To limit the number of users (including teachers) that can be subscribed to any course, globally.

christianbeeznest:
It is added with this PR https://github.com/chamilo/chamilo-lms/pull/6453

Thanks for confirmation.

ywarnier:
The DB changes have been included for alpha 3.
Issue remains open until the feature is implemented into the interfaces.

ywarnier:
Between others, the setting should affect https://github.com/chamilo/chamilo-lms/blob/3b16f4a6990ba472544f7681354f4b48217fb13e/public/main/inc/lib/course.lib.php#L871 (CourseManager::subscribeUser()) to prevent registration of new users, but it should also warn the admins in any course-subscription page, users when a course is available in the catalogue, and teachers when teachers can subscribe users to courses, that *there is* a limit (if there is one), before they subscribe users.

ywarnier:
The setting is in the database, so moving to post-beta phase.

christianbeeznest:
It is checked with this PR https://github.com/chamilo/chamilo-lms/pull/7026

Thanks for confirmation.

ywarnier:
The setting has been added and I see some code uses that limit already, but I do not see it in the interface.
We need to implement it by refusing any action that would add more users:
- [x] in main/user/user.php (registering manually by clicking the "Register" button or by selecting them as a group and using the "Action" dropdown at the bottom of the list)
- [x] this is also true for teachers (they are included in the limit)
- [x] in main/user/class.php when trying to link to a class that would put the total number of users above the limit
- [x] in main/admin/subscribe_user2course.php
- [x] in main/admin/user_import.php if trying to insert users to a course through the "Courses" column in CSV or XML.
- [x] main/admin/course_user_import.php (same)

The idea is that if any operation would pass the limit if executed, the whole operation is cancelled.
So if you have a limit of 5, a course with already 1 teacher and 2 students (so 3 in total), and you try to insert 3 (or more) new users, a message appears:
[...]

PR #7082, opened 2025-11-19
Admin: Enforce global users-per-course limit on course subscriptions - refs #5114

christianbeeznest:


ywarnier:
The code looks good to me.

Issue #2860, opened 2019-03-21
[C2] Error in course voting in catalogue + catalog access


aragonc:
**Describe**
Error al votar un curso en el ranking.

**To Reproduce**
Vote un curso usando las estrellitas del ranking, cuando es Alumno no funciona nada, ni siquiera aparece el error generado en el ajax.

**Screenshots**
![error en accesos](https://user-images.githubusercontent.com/2940509/54782113-206ead00-4bec-11e9-85d0-d9e14f62f857.png)

**Additional context**
Se podra cambiar el tipo de ranking de estrellas por Likes y Dislike?

Zebedeu:
> **Describe**
> Error al votar un curso en el ranking.
> 
> **To Reproduce**
> Vote un curso usando las estrellitas del ranking, cuando es Alumno no funciona nada, ni siquiera aparece el error generado en el ajax.
> 
> **Screenshots**
> ![error en accesos](https://user-images.githubusercontent.com/2940509/54782113-206ead00-4bec-11e9-85d0-d9e14f62f857.png)
> 
> **Additional context**
> Se podra cambiar el tipo de ranking de estrellas por Likes y Dislike?

En mi opinión, no sería bueno dar Dislike en un curso. Sería malo para el autor del curso. Star es una buena forma de evaluación. Debemos encontrar la fuente del error y corregir ...

ywarnier:
Now the issue has changed a bit : votes are working well but, in the catalogue, when I vote, the number of votes is not updated, and when I visit the course, the number of visits is not updated.

<img width="987" height="1077" alt="Image" src="https://github.com/user-attachments/assets/8128e972-a619-4691-9127-1e3fa1dcff7f" />

NicoDucou:
hi @AngelFQC 

The visit count works fine for all.
The vote is working for admin users only. Students or teachers votes are shown when they vote but they are not stored in the DB so when we connect again we only see the vote of the admin user.
There is also access problem. On a new installation from today students and teachers can see the entry "Explore more courses" in the main menu, but when clicking on it they are redirected to the home page of the platform.
The user is redirected to the home page of the platform.

I tried to set "catalog.course_catalog_display_in_home" so that at least I could try with the catalog on the home page, that's why I could see that the student and teachers vote are not stored in the DB, but maybe it's the same access right problem that block the vote from being stored.

ywarnier:
Also add back the info button on courses to see course description (see #7344)

[...]

PR #7085, opened 2025-11-20
Catalogue: Fix voting in catalogue refs #2860

Ilyassbennanii:


ywarnier:
There's a Vue issue in CatalogueCourseCard, where you define a constant allowSelfSignup on line 265 but it already is defined on line 456. This should have broken your "yarn dev" execution.

Can you please review?

When I remove the definition on line 456, the start are not marked anymore when I vote (if I reload the page, my votes appears, but not if I just click on the stars, even if nobody voted before and there's only one vote).

assets/vue/components/course/CatalogueCourseCard.vue
```diff
@@ -227,19 +225,130 @@ const platformConfigStore = usePlatformConfig()
...
+
+    localCourse.value.userVote = { vote: newVote }
+
+    // Emit to the parent to persist (the parent must call the API)
+    emit("rate", { value: newVote, course: props.course })
+  },
+  { immediate: false },
+)
+
+const allowSelfSignup = computed(() => {
```
ywarnier, line 340: This breaks "yarn dev" beause allowSelfSignup is already defined on line 456.
Actually, the one that breaks is the one on line 456, but I mean there's a duplicate definition of the same constant and that does not work.

ywarnier (changes requested):

Issue #7103, opened 2025-11-25
Assignment, add require file format feature.


baelmyhu:
When creating an assignment add the functionality to choose a list a file format allowed for the file the student will upload.

A multi-selection component is used with predefined extensions, and you can add a custom extension via an input text.

Default extension are : pdf, docx, xlsx, zip, mp3, mp4
(@todo on another issue: add in the Chamilo Admin Tool configuration page the possibility to customize the default list)

When upload his assignment, a line show the student allowed file format.
He gots an error message if his file's extension have a wrong format.

<img width="1200" height="431" alt="Image" src="https://github.com/user-attachments/assets/3c07e72d-e9a5-4fed-a380-049833e02ff5" />

ywarnier:
Working well. Will be in beta 2. Thanks!

PR #7104, opened 2025-11-25
#7103 Assignment, add require file format feature

chamilo-uga:
cf issue https://github.com/chamilo/chamilo-lms/issues/7103


ywarnier:
Tested and approved. Thanks!

PR #7120, opened 2025-11-28
Session: Hide remaining days for coaches on my sessions - refs BT#22640

christianbeeznest:


ywarnier:
Merging to include language terms.
Working state not verified but I don't see much harm caused by this and we're still in beta.

PR #7136, opened 2025-12-04
Données corrompues lors des copies des questions MATCHING_COMBINATION…

LudiscapeXApi:
…, MATCHING_DRAGGABLE_COMBINATION

Erreur lors des copies des question sMATCHING_COMBINATION, MATCHING_DRAGGABLE_COMBINATION

### DIFF
--- a/main/exercise/answer.class.php
+++ b/main/exercise/answer.class.php
@@ -1050,7 +1050,7 @@ class Answer
             }
 
             // Fix correct answers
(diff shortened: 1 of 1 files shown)


ywarnier:
Thanks. Well found.
I see this was my mistake (forgot that part in this commit a91286ed33668c5d332398934ae7b89aa8b3b628).
We later added this to 2.0 and apparently Christian saw it and fixed it in 2.0 but not in 1.11.x (6b9facaf80e6139139258a845901cd5be8c70947).
An unlucky combination of events, but difficult to catch.

Issue #3800, opened 2021-02-23
Xapian in v2


jmontoyaa:
Disable it? Update code, use classes? 

Xapian uses the folder:

`api_get_path(SYS_UPLOAD_PATH).'plugins/xapian/searchdb';`

ywarnier:
Details of the issue:
- Xapian has a non-official PHP extension which has changed considerably since the first implementation in Chamilo in ~2008
- The general logic of the code is still the same: there is a configuration section in the Chamilo 2 administration (/admin/settings/search) which has "some" of the parameters needed to define the tags necessary for Xapian to store indexed data (there is probably some work to be done there)
- The search page has to be inserted somewhere flexible inside the Chamilo 2.0 interface. In 1.9/1.10, it was available as a link in the left menu. We probably want to make it more visible (if it is configured and it works), but for now the same logic would work (just a link sending to a search + results page)

Mostly, the issue is about not having maintained this feature over the years, so the code inside Chamilo doesn't match the PHP extension anymore.
Also, the documents tool has been re-written in 2.0, meaning the feature has to be re-introduced.

ywarnier:
In 1.11, the Xapian config page was something like this:
![Capture d’écran du 2023-02-03 14-57-05](https://user-images.githubusercontent.com/609839/216622009-49eb41a2-e90a-4137-b804-ec0946a8e903.png)
![Capture d’écran du 2023-02-03 14-57-24](https://user-images.githubusercontent.com/609839/216622035-0ca22e7c-e76e-423b-86c4-2a190d6e7a52.png)
![Capture d’écran du 2023-02-03 14-57-46](https://user-images.githubusercontent.com/609839/216622043-06f5ba5a-62ba-4eca-ae07-c10eb6a1c2c6.png)


But in 2.0 it has been reduced (and mixed) like this:
![Capture d’écran du 2023-02-03 14-55-42](https://user-images.githubusercontent.com/609839/216622106-3d275a92-f77d-427d-b70e-928a5ffaf738.png)

So there is non-anecdotical time to be spent on how to restore the right settings (more than anything the "Add specific search fields" section) and show the right info regarding the suggested converter libraries.

ywarnier:
I have no idea what that "Number of upcoming events" has to do in the search section. I think it must have been a typo at the time of splitting config settings (more sections in 2.0 than there were in 1.11).

ywarnier:
Clearer, updated list of to-do for this issue:
[...]

PR #7144, opened 2025-12-08
Search: enable Xapian indexing for course resources - refs #3800

christianbeeznest:


christianbeeznest:
I close this PR to avoid multiple conflicts with other version, will be sent other PR clean

PR #6896, opened 2025-10-10
Fix space available widget - ref #6719

Coussecousse:
Hi !

I refactored the code and implemented the space chart based on Chamilo v1.
I also added the ability to view other teachers’ spaces (it just felt right somehow).
I’m not sure if it works with groups, since I couldn’t find where to test that — hopefully someone can check! Theoretically, it should :laughing:

I hope my code is fine !

**Issue** : https://github.com/chamilo/chamilo-lms/issues/6719

**Screens :** 
Capture d’écran du 2025-10-10 15-35-30

**If error :** 
Capture d’écran du 2025-10-10 15-30-56
(It is centered, I fixed it in the commit `Minor` — I just took the screenshot before noticing it wasn’t.)

:warning: To test sessions, I had to create one and try it through another teacher. I noticed it didn’t work at first, so I fixed it — I’ll create another PR for that.

Issue #6719, opened 2025-09-09
[C2] Fix space available widget


ywarnier:
The new document tool doesn't support space usage chart yet. Fix it (check 1.11 for guidance).

Currently, it looks like this

<img width="1574" height="777" alt="Image" src="https://github.com/user-attachments/assets/5291542a-9b57-44a9-a4fb-3846e176d582" />

ywarnier:
Perfect. Thanks to you both @Coussecousse and @christianbeeznest !

PR #7178, opened 2025-12-16
Fix space available widget - ref #6719

Coussecousse:
Original pull request : https://github.com/chamilo/chamilo-lms/pull/6896

Hi! Sorry this is taking sooo long!

I looked at the code and your reviews @ywarnier, and I did the following : 
- Removed the ECS fixe typo commit
- Used the course entity to get the sessions and retrieve the group
I didn’t manage to use `CidReqHelper` because the request does get the `courseId` and the `course` entity, but not `sid` and `gid`.

Dump of the request -> you can see `cid` and `course` : 
Capture d’écran du 2025-12-16 14-49-59
Differents dumps : 
- Line 35 : The course
- Line 2471 and 37 : api_get_session_id()
- Line 43: Just again the request from `CidReqHelper`
- Line 54 : `sessionId` using `$this->cidReqHelper->getSessionId();` 
Capture d’écran du 2025-12-16 14-50-20

So if the `session` is null, then I can't know what size is used (and the calcul crash)...
 
However, using the course entity still allows the feature to work properly.
Capture d’écran du 2025-12-16 15-34-19

I’ll let you review this hopping there's not much to correct :D !

### DIFF
--- a/src/CoreBundle/Controller/Api/DocumentUsageAction.php
+++ b/src/CoreBundle/Controller/Api/DocumentUsageAction.php
@@ -4,132 +4,107 @@ declare(strict_types=1);
 
 namespace Chamilo\CoreBundle\Controller\Api;
 
-use Chamilo\CoreBundle\Helpers\CidReqHelper;
 use Chamilo\CoreBundle\Repository\Node\CourseRepository;
 use Chamilo\CourseBundle\Repository\CDocumentRepository;
-use Chamilo\CourseBundle\Repository\CGroupRepository;
-use DocumentManager;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpKernel\Attribute\AsController;
 
-
 #[AsController]
-class DocumentUsageAction extends AbstractController
+final class DocumentUsageAction extends AbstractController
 {
+    /**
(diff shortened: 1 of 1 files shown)


ywarnier:
Hi @Coussecousse 
No problem about the timing, we all have many things to do and this is not a priority for right now (but we have to close it before the stable).

I notice in [CDocumentRepository.php](https://github.com/chamilo/chamilo-lms/pull/7178/files#diff-2c04e1b7f1c85414d5beb121658dbc88bc8c90e1ba554fb78f753bafe60cc8dd) you're still using a path comparison.

We did a change recently which affects considerably the way we work with documents, in that the c_document table (mapped by the CDocument entity) is not the main way to list documents anymore. Rather, the resource_link table has been modified to include a "parent_id", which helps reflect the hierarchical structure of the documents in any course context (because we intend on making documents shareable between courses and we still need them to maintain a local-context hierarchy, which we couldn't do with c_document anymore).

This means that the right way to get the "root" (the equivalent of the "/" path) is not to check for "/" in the path anymore, but rather to check for the "eldest parent" in the resource_link hierarchy of nodes, or (to be more efficient) the direct descendant of the document "tool" (which is also a node of the course).

I'm not entirely knowledgeable on the way to manage this in Chamilo right now (this is a recent change), but I'm adding @christianbeeznest here to review that part of your PR. The rest seems fairly reasonable and your explanation about the sid is satisfying to me (maybe Christian will add something to this, but I think it's fine).

PR #7368, opened 2026-01-14
Fix E_NOTICE in api_htmlentities function about deprecation with mbstring

AngelFQC:


ywarnier:
Maybe this would work with Chamilo 1.11 for PHP 8.3 as well.

PR #7392, opened 2026-01-18
Internal: Migrate display.show_tabs from legacy CSV to JSON

christianbeeznest:


ywarnier:
Tested and approved.
This is the setting before the change:
Capture d’écran du 2026-01-18 22-51-15

This is the setting after the change:
Capture d’écran du 2026-01-18 23-01-03

This is the menu for admins with all options by default (but some features already enabled in comparison with a default install):
Capture d’écran du 2026-01-18 23-03-04

This is the same menu after setting diagnostis, search and question_manager to false:
Capture d’écran du 2026-01-18 23-02-52

PR #7393, opened 2026-01-18
Internal: Remove deprecated tool_visible_by_default_at_creation setting

christianbeeznest:


ywarnier:
Works for me.
Before the change:
Capture d’écran du 2026-01-19 00-17-44

After the change:
Capture d’écran du 2026-01-19 00-18-03

ywarnier:
When unselecting 2 tools from the course.active_tools_on_create, I can indeed see a new course has those 2 tools disabled upon creation.
Capture d’écran du 2026-01-19 00-20-37
Capture d’écran du 2026-01-19 00-20-27

Issue #6047, opened 2025-01-24
[C2] CSS class per global page type


ywarnier:
In order to customise the appearance of Chamilo, we need to introduce a CSS marker (a div with a specific CSS class) at the top of each page's content.

I can see the following pages:
- page-home 
- page-my-courses (/courses)
- page-catalogue (/catalogue/courses)
- page-agenda
- page-tracking (with all sub-tracking pages under the same name)
- page-social (in general)
- page-account-security
- page-videoconference
- page-diagnosis
- page-administration
- page-administration-user (and other pages of that block)
- page-administration-course (and other pages of that block)
- page-administration-user (and other pages of that block)
- page-administration-session (and other pages of that block)
- page-administration-platform (and other pages of that block)

And optional:
- page-tool-document
- page-tool-exercise
- ...

The fact that it's legacy or not doesn't really matter, as long as the CSS class is placed correctly.

christianbeeznest:
It is checked with this PR https://github.com/chamilo/chamilo-lms/pull/7430

Thanks for confirmation.

ywarnier:
Fixed.

PR #7430, opened 2026-01-25
Internal: Add global page-type CSS markers (SPA + legacy) - refs #6047

christianbeeznest:


ywarnier:
Checked and approved. The questions admin page (main/admin/questions.php) shows `page-administration page-administration-platform`. It's not a big deal, but it is exactly the same as the generic /admin page, for some reason.
I would have expected `page-administration page-administration-question` or something like that.

PR #7477, opened 2026-02-01
Internal: Migration: prevent document duplicates and improve file link rewrite

christianbeeznest:


NicoDucou:
@christianbeeznest 
all good with this PR, I can not accept it because there are conflicts now and I'm not sure which version to keep.
So please fix the conflicts and you can merge the PR.

Issue #7530, opened 2026-02-17
[C2] Implement global calendar features


ywarnier:
The global calendar still lacks a few features that were present in 1.11.

<img width="1425" height="922" alt="Image" src="https://github.com/user-attachments/assets/85c7fe7e-a2c8-4ae6-882f-87cf630ad6d6" />

Namely
- Add a new event: shows a modal or redirects to a form. The basic version would be to only make personal events. The more advanced view would be to offer a selection of personal, course, course/session or global event creation, but that will depend on roles and permissions.
- Sessions plan calendar: a flat view of the 52 weeks of the year with one subscribed session per line, organized chronologically)
- My students schedule: a view that shows the events - including assignments handing over dates - of the students of a given session. I need to select a session where I am a tutor for that to work, but once I am a tutor in a session, I can see *all* events in that session, not only (and this is an exception to the normal use case) the events of the courses I am a tutor of.

An idea of the expected appearance of the sessions calendar: 

<img width="1043" height="577" alt="Image" src="https://github.com/user-attachments/assets/8e5e03b5-01d0-444b-b9d1-d15bc9b5c8f3" />

The other interfaces are relatively obvious (one is a form to create an event - we already have this form when clicking on the calendar to open a new even - , and the other is just the calendar view with a special filter)

christianbeeznest:
There is a version about these features in this PR https://github.com/chamilo/chamilo-lms/pull/7571

Thanks for confirmation.

christianbeeznest:
It is checked and improved with this PR https://github.com/chamilo/chamilo-lms/pull/7571 , merged with master now. 

Thanks in advance for your review.

ywarnier:
Yup, that's much better. Only *one* detail : in the popup view for the session event in another course (where it says "You cannot access the linked course from this view."), the popup should include the title of the course. Otherwise it's really difficult to sort through all the events for a teacher in another course.

ywarnier:
Just for later, the sessions plan calendar now shows correctly:

<img width="1615" height="492" alt="Image" src="https://github.com/user-attachments/assets/1f9d784a-815d-409b-8fec-4002e8cb5dc0" />

christianbeeznest:
> Yup, that's much better. Only _one_ detail : in the popup view for the session event in another course (where it says "You cannot access the linked course from this view."), the popup should include the title of the course. Otherwise it's really difficult to sort through all the events for a teacher in another course.

[...]

PR #7571, opened 2026-02-23
Agenda: Add session plan and calendar schedule views - refs #7530

christianbeeznest:


ywarnier:
This PR has a mix of other things in there. It should be fixed (icons file, locales files, etc).
When setting up the right sessions (3 sessions that span january+february, then february+march, then march+april), the student gets something like this, which is incorrect:

Capture d’écran du 2026-02-25 01-57-32

The numbers in the columns are *weeks*, so if a session spans january+february, it should go roughly from week 1 to week 8. In the screenshot, we can see it's only a fraction of week 1.

The schedule view for a session doesn't do what I am expecting either.

Here are the testing procedures.

## Sessions plan
- [ ] Create a course A
- [ ] Create session 1 from 1/1 to 28/2, with course A in it and assign student acostea
- [ ] Create session 2 from 15/2 to 30/3, with course A in it and assign student acostea
- [ ] Login as acostea
- [ ] Go to calendar
- [ ] Click on sessions plan icon
- [ ] The calendar plan appears (52 columns) with session 1 spanning from week 1 to week 8 or 9, and session 2 from week ~6 to week ~12

## "My students schedule"

- [ ] Create courses A, B and C
- [ ] Create session 3 with courses A, B and C in it. Session 3 should span from 1/2 to 31/3.
- [ ] Assign session to student acostea
- [ ] Assign different teachers to each course in this session
- [ ] As admin, go to course A in session 3, create an event "abc" on the 15/2
- [ ] As admin, go to course B in session 3, create an event 'bcd" on the 18/2
- [ ] Login as the tutor of the course C in session 3
- [ ] I should not see courses A and B in the session block in "My sessions" (these are not *my* courses)
- [ ] Enter the global calendar
- [ ] Click "My students schedule" icon
- [ ] I should see events "abc" and "bcd" appear in february
- [ ] I should not be allowed to enter the courses linked to the events; only see the event and the details of the event.

christianbeeznest:
> This PR has a mix of other things in there. It should be fixed (icons file, locales files, etc). When setting up the right sessions (3 sessions that span january+february, then february+march, then march+april), the student gets something like this, which is incorrect:
> 
> The numbers in the columns are _weeks_, so if a session spans january+february, it should go roughly from week 1 to week 8. In the screenshot, we can see it's only a fraction of week 1.
> 
> The schedule view for a session doesn't do what I am expecting either.
> 
> Here are the testing procedures.
> 
> ## Sessions plan
> 
> ## "My students schedule"

Hi Yannick, pushed a follow-up commit that:

cleans up the PR by removing/splitting unrelated changes (icons/locales),

fixes the Sessions plan week-span for sessions crossing months (Jan–Feb now shows ~week 1→8/9),

updates “My students schedule” to show the Feb events but prevent course access from those links (and tutors only see their own courses in “My sessions”).

PR #7704, opened 2026-03-06
Install: Fix missing default value for not null field qualify - avoid…

ywarnier:
… install error message
Not completely convinced we should do this. Still thinking.


ywarnier:
We decided to drop this as we could not reproduce the initial reported error.

PR #8018, opened 2026-03-15
Webservice: Move tracking stats endpoints to /api/tracking - refs #23304

christianbeeznest:


ywarnier:
@christianbeeznest please review and fix the PHP format code & PHP static analysis reports above.

PR #8095, opened 2026-03-30
Update trad4all.inc.php

moigaston2:


ywarnier:
Hi @moigaston2 
I'm not against the Fähigkeit to Kompetenz change, but the "Badges" are different from "Skills". They are visual representations of skills, not the skills themselves so the removal of the "Fähigkeit" term completely does not seem acurate.

PR #8154, opened 2026-04-07
Add service management for Chamilo limits changes

ywarnier:
This is untested.
Should merge into 2.1.


ywarnier:
So this has been implemented in #8188, but I'm adding #8351 to complete this with subscription-type payments

PR #8193, opened 2026-04-15
Internals: make email verification compliant with RFC6531 to accept s…

NicoDucou:
…pecial caracters - refs BT#22928


ywarnier:
This one looks good to me @NicoDucou 
Any reason to hold it?

NicoDucou:
I did not merge it yet because not all SMTP implement the use of special caracters so it may generate an error in some SMTP and the mail will not be delivered.

ywarnier:
Left "on hold" in case someone experiences an issue in the future.

PR #8202, opened 2026-04-18
Document: Restore document.users_copy_files setting in modern Documents and Personal Files

christianbeeznest:


src/CourseBundle/Entity/CDocument.php
```diff
@@ -236,6 +237,16 @@
             security: "is_granted('ROLE_USER')",
             deserialize: false
         ),
```
AngelFQC, line 239: Operation moved to PersonalFile entity `/api/documents/{document_id}/personal_files`

Now it returns a PersonalFile object
cf625d1d65b44724968ef422fcd847cb8025c442 eacfd9c84d8124e59647b3c92ece9d7c4bbc466e

PR #8213, opened 2026-04-20
Re-enforce https checking in reverse proxy cases

ywarnier:
Add an additional layer of check for https URLs in learning paths lists and other pages.


ywarnier:
In the end, this didn't seem necessary. The initial problem was coming from an improper configuration of `TRUSTED_PROXIES` in `.env`

PR #8360, opened 2026-05-05
Cron: Fix leading zeros stripped from Matricule/official_code on XLSX…

ywarnier:
… import - refs BT#23385


ywarnier:
Abandonned (does not seem to be the right way to solve the human behaviour behind the reported issue).

PR #8369, opened 2026-05-07
CI: Improve speed, extend and fix createUsers Behat test

ywarnier:


ywarnier:
Abandoned as the gains were not very significant *and* featureContext has been worked on separately by someone else in #8656 *and* we are migrating to Playwright.

PR #8653, opened 2026-06-22
Security: Remove path-traversal-prone placeholder endpoints in FileManagerController

AngelFQC:
## Summary

`FileManagerController` shipped a set of placeholder/scaffolding actions, none of which are referenced anywhere in the codebase (verified across `src/`, `public/`, `assets/`, `templates/`, `config/`, `tests/`, and route-name / URL / `generateUrl` lookups). One of them, `download()`, built the served file path directly from the `{filename}` route parameter and handed it to `BinaryFileResponse`, exposing a path-traversal sink reachable by any authenticated user.

Refs GHSA-7jqm-3829-36cm

## Fix

The unused placeholder actions — `download`, `list`, `upload`, `createFolder`, `rename`, `delete` — are removed, which deletes the path-traversal sink at the source rather than merely containing it. The now-unused `BinaryFileResponse` import is dropped.

Only `uploadImage` is kept: it has a real implementation backed by `PersonalFile` / `BaseResourceFileAction`.

## Invariant now enforced

There is no longer any route in this controller that turns client input into a filesystem path. The traversal primitive is gone, not just sanitized.

## Note

The remaining `uploadImage` action is also currently unreferenced; full removal of the controller is under internal review by the team and can follow separately if confirmed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/CoreBundle/Controller/FileManagerController.php
+++ b/src/CoreBundle/Controller/FileManagerController.php
@@ -106,6 +106,10 @@ class FileManagerController extends AbstractController
     #[Route('/download/{filename}', name: 'file_manager_download', methods: ['GET'])]
     public function download(string $filename): Response
     {
+        // Strip any directory component so the route parameter cannot be used to
+        // traverse outside the intended storage directory (e.g. "..%2F..%2Fetc%2Fpasswd").
+        $filename = basename($filename);
+
         // Implement logic to download files
         // Replace 'path/to/your/files' with the actual path where the files are stored
         $filePath = 'path/to/your/files/'.$filename;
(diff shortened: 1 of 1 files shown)


AngelFQC:
Updated the approach: instead of containing the traversal in `download()` with `basename()`, the unused placeholder actions (`download`, `list`, `upload`, `createFolder`, `rename`, `delete`) are removed outright, eliminating the path-traversal sink at the source. `uploadImage` is kept since it has a functional implementation. Full removal of the controller is being reviewed internally and can follow if confirmed.

PR #8667, opened 2026-06-27
Learnpath: Migrate legacy code to vue + Api symfony

christianbeeznest:


ywarnier:
- [x] migration file should be in Schema/V210/
- [x] lists of languages to select from (in the LP creation process for example) shows languages without proper sort order
- [x] In the LP builder view, the "Parent" area of any new document should be set to the learning path itself, by default, otherwise everyone will leave it empty.
- [x] when *creating* a new document from the LP builder view for the first time, the "Destination folder" says I have "No available options", but if I don't select any, then the document fails to create and I get an error message inside the form (and I have no idea, as a standard user, how to solve this). In the code, there should be something create a folder inside the Documents tool called "Learning paths" (consistent with previous behaviour). I had to go in the Documents tool, create a folder and a subfolder to recreate previous behavior.
- [x] When the folder already exists and when I create a new document, the "Destination folder" should be preset to the folder of that learning path
- [x] When using the "Upload" option for documents, the Destination folder should try to pre-select the folder of the current learning path
- [x] When creating my own learning path from Chamilo with 4 html documents, when I preview it as a teacher, opening the first step should be enough for me to see it progress to 25% (because it is the equivalent of an asset). I should not need to do anything else than open the document before it is set to completed. For now, progress stays at 0% all along. It seems that adding prerequisites might unlock that progress (I'm not sure, but now I have progress and I added prerequisites for all pages a few minutes ago), but progress should not be dependent on prerequisites.
- [x] as admin or teacher, when taking a SCORM 1.2 learning path, it generally works, but when I click on the "Reporting" icon in the "Home" menu inside the learning path, I get redirected to something like /resources/lp/132/4/reporting?cid=2&origin=learnpath&isStudentView=false&gradebook=0 where I can see the standard Chamilo menu on the left and I cannot see my own progress or score. It looks like it's trying to give me a list of different users that have taken this learning path, while all I should see is *my own result* for this particular attempt. See v1.11 or the legacy version of LP for a clear example. It should *not* be the same screen/report as the "Reports" icon I can see from the learning paths list! (this one looks correct).
- [x] each time I clic on a learning path (from the list of learning paths) I see the screen refreshing and the standard menu on the left and the header appearing before another refresh that makes the same menu disappear and I see the full re-appearance of the learning path viewer. Eliminate the intermediate refresh, if possible
- [x] the icon for tests in the left menu of the LP should be the same icon as the test tool icon (not a plain check-mark as the current one)
- [x] there's no completion symbol for tests in a Chamilo LP either (see previous point 4 items above here), so I'm not sure whether it would appear correctly visually once the progress works.
- [x] not sure it's related but when I create a link (from the learning path, but it sends me to the links tool), the default language in advanced setting should be set to the language of the course by default (if we're in a course context). 
- [x] I erroneously clicked "Edit" on a SCORM 1.2 learning path. There, I could add (apparently) all the types of LP items, but in fact I can create items but I cannot drag & drop them inside the SCORM. Editing a SCORM is a delicate operation, so I would prefer *not* to find the edit icon for any SCORM package (currently the edit icon exists but, for example, assignments and links *can be created* but not inserted inside the learning path). When I do create items, I often get the following JSON error at the end of the creation (apparently it tries to send me back to the learning path with the item inserted in the LP structure but something prevents it): `The structure of an imported learning path cannot be changed here.`. I can then come back to the learning path and the item has been created but I cannot add it to the table of contents (which is fine). Forum is added without issue though, which is more confusing. A SCORM-type LP should not show the pencil icon as the edition is made outside of Chamilo.
- [x] when I try creating a forum from the learning path, the redirect works (maybe *because* it opens in a modal instead of sending me to the forum tool). In the forum creation modal, I see 2 "Cancel" buttons. I didn't need them, but obviously this is slightly confusing.
- [x] Moving items between no-chapter and a chapter make the right side of the screen repaint itself. Not ideal but not blocking either.
- [ ] the last item special page for certificates loads when there is normal progress accounting in the LP, but I get "The file was not found". Maybe because it points to some legacy page?
- [x] At first glance, the admin part of assigning users to a LP seems to work, but the "Edit" (pencil) icon next to each user's name doesn't seem apropriate. It should be an addition (green button with a +) instead of a pencil. The pencil is ok to edit someone already subscribed to the LP, but not if the person is not subscribed yet. The date restrictions seem to work fine but the date selector does not allow to pick a time (you have to write it). Make it possible to choose the time (not only the date) with the datepicker. 
- [x] In general, the simple editor view for the LP seems to load a bit slowlier than the previous version, but maybe it's a caching issue on my side.

ywarnier:
Merged by mistake (uups). Will have to work quickly on those improvements :sweat_smile:

ywarnier:
I'm officially removing support for AICC HACP as part of this effort (commit 8802ce3e0ca).

HACP is now insecure. The core issues:

- Session ID as sole auth: HACP communication is authenticated only by an AICC_SID token passed as a plain HTTP parameter/POST field. Anyone who captures that value (via network sniffing, proxy logs, browser history, or a leaked Referer header) can replay it and read or forge completion/score data for that session.
- No mandated encryption: The spec (from 1998) predates any expectation of TLS. Many legacy implementations run it over plain HTTP by default; even over HTTPS, the protocol itself has no additional integrity or replay protection layered on top.
- No CSRF/replay protection: It's just simple HTTP POST/GET exchanges with no nonce, signature, or expiry beyond whatever the LMS chooses to bolt on.
- Course-package-controlled origin: The HACP host/URL is typically embedded inside the AICC content package launch parameters, which is a soft trust boundary — a tampered package can point communication elsewhere.

christianbeeznest:
All the post-merge observations reported here have now been addressed in a follow-up PR https://github.com/chamilo/chamilo-lms/pull/8724 , which has already been merged into master.

The fixes cover the V210 migration location, language sorting, LP document folders and defaults, native document progress, SCORM runtime reporting and editing restrictions, SPA navigation, Tests icons and completion status, course-language defaults, forum modal actions, certificate handling, advanced access icons and date-time selection.

The affected functional scenarios were retested after the follow-up merge. The Learning Path migration and its urgent corrections are now available in master.

ywarnier:
- [x] Improve the style of the reporting page inside the LP: It should appear as a frame with the LP ToC column on the left (if it is set to visible), and there should be more spacing between the border of the page and the title of the page, title of the learning page and "Switch to student view" button (they are too close to the borders)
- [x] The Xapian indexing checkbox does not appear in the LP creation form, but it does in the update form (add it to the advanced parameters in the creation form)
- [x] The language should be set to "No specific language" by default. If I use this option and then I later edit the learning path config, the language is empty again.
- [x] When viewing a LP, then exiting to the course homepage for the first time in a course, I am switched to student view. If I click on the LP tool and see the LP list, I am still in student view and invisible learning paths do not appear. When I click the "Switch to teacher view" button, the hidden learning paths do not appear again. I have to then refresh the page (F5) for the hidden learning paths to appear, despite being in teacher view. This seems to only work once before we have switched back to teacher view, so I think you need to create a new course first. Not sure. Once you have clicked on "Switch to teacher view", the issue does not appear in any subsequent request.
- [x] In my previous review, I mentioned SCORM LPs should not be editable, but Nico reminded me that they should in fact be editable, so this change needs to be reverted and the broken actions need to be repaired. The edition of a SCORM is not recommended (documents from the SCORM should not be editable) but you can still add new documents within the SCORM. The prerequisites will then not work correctly, though.
- [x] The certificate "end of LP" page does not work yet. This is the only point that doesn't seem to have been fixed.
- [x] In advanced-access, the "+" icon on the green button does not appear (see screenshot). Probably the wrong icon name used.

Capture d’écran du 2026-07-14 11-16-16

christianbeeznest:
All reported Learning Path post-migration issues have been addressed in the PR https://github.com/chamilo/chamilo-lms/pull/8732 , which has now been merged into master. Reporting is displayed inside the LP player, the Xapian option is available in advanced settings, teacher view is restored after leaving the player, SCORM editing was restored, the final certificate item was fixed, and the Advanced access and header actions were corrected. No translation or database changes were included.

ywarnier:
This is to confirm that all is well with this PR (and extended commits) now. Thanks. Great work!

PR #8672, opened 2026-06-29
Fix: allow ROLE_ADMIN to POST /api/course_rel_users for other users

addow:
## Summary

- `securityPostDenormalize` on the POST operation lacked an admin bypass: admins received 403 when subscribing any user other than themselves
- The upstream security commit (2a9f060fa9) enforces self-enrollment but omitted `ROLE_ADMIN` from the post-denormalize check
- Mirrors the `ROLE_ADMIN` bypass already present in `CourseRelUserStateProcessor`

## Test plan

- [ ] POST /api/course_rel_users as ROLE_ADMIN with `user` = another user → HTTP 201
- [ ] POST /api/course_rel_users as regular user with `user` = self → HTTP 201
- [ ] POST /api/course_rel_users as regular user with `user` = other user → HTTP 403

### DIFF
--- a/src/CoreBundle/Entity/CourseRelUser.php
+++ b/src/CoreBundle/Entity/CourseRelUser.php
@@ -39,7 +39,7 @@ use Symfony\Component\Validator\Constraints as Assert;
         ),
         new Post(
             security: "is_granted('ROLE_USER')",
-            securityPostDenormalize: 'object.getUser() == user',
+            securityPostDenormalize: "is_granted('ROLE_ADMIN') or object.getUser() == user",
             processor: CourseRelUserStateProcessor::class
         ),
         new GetCollection(


AngelFQC:
Thanks @addow — you're right that admins should be able to enroll others. But the inline bypass was too narrow and masked two pre-existing gaps in the same Post: the owner check is the only guard against enrolling third parties (the processor only validates which course, not who), and status was mass-assignable (a self-enroller could POST status: TEACHER and escalate). We've folded this into a fuller fix — a CourseRelUserVoter (CREATE/VIEW) centralizing the rules (admins/course-teachers enroll anyone, others only themselves into catalogue courses) plus a processor forcing a safe status, with regression tests. Closing in favor of daf5e026069e8e04a4e3acf55a5bda1bea5ccf37 — thanks for surfacing the admin case!

PR #8675, opened 2026-06-29
Course: add Delete operation to /api/courses/{id} (admin only)

addow:
## Summary

Adds `DELETE /api/courses/{id}` to the Course `#[ApiResource]`, restricted to `ROLE_ADMIN`.

- Uses API Platform's built-in remove processor (`EntityManager::remove()` + flush)
- The database-level `onDelete: CASCADE` on `resource_node_id` handles cleanup of the linked ResourceNode
- Mirrors the pattern used by other destructive operations on Course (`Patch`) and Session (`Delete`)

## Notes

`CourseVoter::DELETE` grants course teachers delete rights, which would be too permissive for a course-level delete. The explicit `is_granted('ROLE_ADMIN')` sidesteps the voter and restricts access correctly.

Physical files (e.g. course illustrations on disk) are not automatically removed on delete — this is a pre-existing limitation shared by the legacy `CourseRepository::deleteCourse()`.

### DIFF
--- a/src/CoreBundle/Entity/Course.php
+++ b/src/CoreBundle/Entity/Course.php
@@ -11,6 +11,7 @@ use ApiPlatform\Doctrine\Orm\Filter\SearchFilter;
 use ApiPlatform\Metadata\ApiFilter;
 use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\Delete;
 use ApiPlatform\Metadata\Get;
 use ApiPlatform\Metadata\GetCollection;
 use ApiPlatform\Metadata\Post;
@@ -42,6 +43,7 @@ use const SORT_NATURAL;
     operations: [
         new Get(security: "is_granted('VIEW', object)"),
         new Post(security: "is_granted('ROLE_TEACHER') or is_granted('ROLE_ADMIN')"),
+        new Delete(security: "is_granted('ROLE_ADMIN')"),
         new GetCollection(
             paginationClientEnabled: true,
             security: "is_granted('ROLE_TEACHER') or is_granted('ROLE_ADMIN')"


addow:
## Follow-up: filesystem and ResourceNode cleanup are not handled

After further investigation, I want to flag a limitation of this implementation before it is merged.

**What happens when `DELETE /api/courses/{id}` is called:**

API Platform's default remove processor calls `EntityManager::remove($course)` + `flush()`. This only deletes the `course` DB row. The following are **not** cleaned up:

| Left behind | Root cause |
|-------------|-----------|
| `resource_node` row (and all child rows) | The FK is on `course.resource_node_id → resource_node.id`. `onDelete: CASCADE` means "delete course when resource_node is deleted" — not the reverse. `AbstractResource` only has `cascade: ['persist']`, not `cascade: ['remove']`, so Doctrine never schedules the ResourceNode for removal. |
| All `resource_file` rows | `ResourceNode.resourceFiles` has `cascade: ['persist', 'remove']`, but this only fires if the ResourceNode is explicitly removed via Doctrine — which it isn't. |
| Physical files in `var/upload/resource/` | VichUploader's `RemoveListener` only fires when a `#[Vich\Uploadable]` entity (`ResourceFile`) goes through Doctrine's `preRemove`. Since no `ResourceFile` is ever Doctrine-removed, no files are deleted from disk. |

**The correct approach** would be a custom `StateProcessor` for the Delete operation that walks the ResourceNode tree, removes each `ResourceFile` via Doctrine (so VichUploader fires), and then removes the ResourceNode — before the Course row itself is deleted. `IllustrationRepository::deleteIllustration()` already demonstrates this pattern for the illustration sub-resource.

The commented-out node-walking code in `CourseRepository::deleteCourse()` appears to be a scar from a previously abandoned attempt at the same problem.

**Options going forward:**
1. Merge as-is with the understanding that orphaned files need manual cleanup (same behaviour as the legacy `CourseRepository::deleteCourse()`)
2. Add a custom `CourseDeleteProcessor` that properly cleans up the ResourceNode hierarchy before deletion

Happy to implement option 2 if that is the preferred direction.

AngelFQC:
Thanks for this. We're going to hold off on exposing course deletion through the API for now — there are currently no API clients that delete courses, so it's not needed.

Deletion is handled through the controllers (`CourseListActionController` and `CourseMaintenanceController`), which now route through `CourseHelper::deleteCourse()`. That path performs all the required cleanup (users, sessions, tracking, tickets, skills, sequence-resource guard, plugin hooks, Xapian purge, system log) and respects multi-URL semantics. The built-in remove processor used here does none of that, so we prefer to keep deletion on the controller path. Closing for now.

PR #8685, opened 2026-07-07
Internal: Fix language flipping between SPA and legacy pages + tests for CourseFromRequestHelper

AngelFQC:
## Problem

With a platform in Spanish, a user profile in English and courses in Spanish, navigating **course → administration → BuyCourses plugin → my courses → administration** kept flipping the interface language on every hop.

## Root cause

The Vue app resolved its locale independently from the backend and persisted it in `localStorage`:

- `i18n.js` booted from `localStorage.app_locale` **before** the server-provided `<html data-lang>`, so a stale value from a previous course visit overrode the locale `LocaleSubscriber` had just resolved for the current request. Legacy pages (BuyCourses) always follow the server, so SPA and legacy pages permanently disagreed.
- `useLocale()` only recomputed the locale when `language_priority_*` settings were configured — without them, leaving a course through client-side navigation kept the course language.
- `App.vue` had two competing locale watchers (one of them watching a non-existent `user.language` field — the entity exposes `locale`).

The initially suspected `$cidReset` guard in `LocaleSubscriber` was verified to be a no-op for this bug: BuyCourses URLs carry no `cid`/`cidReq`, so course resolution already returns null there.

## Fix

- **`assets/vue/i18n.js`**: boot from `<html data-lang>` — the server answer already factors platform/user/course settings, priorities, `?_locale` and `Accept-Language`. Drop `localStorage` persistence entirely (it can only agree with the server or be stale).
- **`assets/vue/composables/locale.js`**: `useLocale()` now mirrors `LocaleSubscriber::getCurrentLanguage()`: `language_priority_1..4` first, then the same last-non-empty fallback (course > user profile > platform). `course_lang` is only computed when a course is in context.
- **`assets/vue/App.vue`**: single locale writer — one `watch(appLocale) → setLocale()`.

Runtime language switching is preserved: entering a course through client-side navigation still switches the interface to the course language instantly.

Also includes the PHPUnit test for `CourseFromRequestHelper` (complements 35060708bf5, which introduced the helper without its test): locks the shared cid/sid/gid resolution contract — numeric id or course code, legacy 1.11.x `cidReq`/`id_session`/`gidReq` fallbacks, attribute-over-query precedence, zero/empty/non-scalar meaning “no context”.

## Testing

- Headless-Chrome (CDP) run of the reported cycle with user `en_US`, platform `es`, course `es`, BuyCourses enabled: every hop (full loads and client-side navigations) now shows the same language the server would render, and `localStorage` is never written. The same script against the previous code reproduces the reported flipping (stays in Spanish when leaving the course client-side, writes `app_locale=es` which poisons the next full load).
- `CourseFromRequestHelperTest`: 19 tests / 52 assertions, green.
- ESLint/Prettier: no new issues.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/assets/vue/App.vue
+++ b/assets/vue/App.vue
@@ -324,20 +324,16 @@ api.interceptors.response.use(
 
 platformConfigurationStore.initialize()
 
-// i18n sync
-watch(
-  () => route.params,
-  () => {
-    const { appLocale } = useLocale()
-    if (appLocale?.value && locale.value !== appLocale.value) setLocale(appLocale.value)
-  },
-  { immediate: true },
-)
+// i18n sync — single writer. appLocale mirrors the server-side locale chain
+// (see useLocale) and reacts to store changes (platform config, user profile,
+// course context set/cleared by the router guards) on client-side navigation.
+// The boot locale comes from <html data-lang>, already resolved by the server.
+const { appLocale } = useLocale()
 
 watch(
-  () => securityStore.user?.language,
-  (lang) => {
-    if (lang && locale.value !== lang) setLocale(lang)
+  appLocale,
+  (newLocale) => {
+    if (newLocale && locale.value !== newLocale) setLocale(newLocale)
   },
   { immediate: true },
 )
--- a/assets/vue/composables/locale.js
+++ b/assets/vue/composables/locale.js
@@ -17,6 +17,11 @@ export function useLocale() {
 
   const appLocale = ref(document.documentElement.dataset.lang)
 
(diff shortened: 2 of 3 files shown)


ywarnier:
Works well for me, thanks!

ywarnier:
I tested with a French portal and one (admin) user in English, one (teacher) user in Spanish, and courses in English and Spanish. It used to switch frequently between languages. Now it behaves as expected (returns to the user language out of course contexts, and with the default platform language when not logged in)

PR #8724, opened 2026-07-13
Learnpath: Fix post-merge builder and runtime issues

christianbeeznest:


christianbeeznest:
## Learnpath post-merge verification

- [x] Migration `Version20260702120000` exists only under `Schema/V210`.
- [x] Language selectors are sorted by the labels displayed to users.
- [x] New documents default `Parent` to the current learning path.
- [x] The first document creation automatically creates `Documents/Learning paths/<LP title>`.
- [x] New documents preselect the existing folder of the current learning path.
- [x] Document uploads preselect the folder of the current learning path.
- [x] Four native HTML documents progress to 25%, 50%, 75% and 100% by opening them.
- [x] Native HTML completion does not depend on prerequisites.
- [x] Runtime Reporting for SCORM shows only the current user's attempt, score and progress.
- [x] The Reports action from the LP list still shows the administrative multi-user report.
- [x] Opening an LP from row and card views does not briefly render the standard course layout.
- [x] Quiz items use the same icon as the Tests tool.
- [x] Completed quiz items show a separate completion indicator and update LP progress.
- [x] Links created from the builder default to the course language.
- [x] Imported SCORM packages do not expose the structure-editing pencil.
- [x] SCORM Update and CStudio-specific editing remain available where applicable.
- [x] The forum creation modal contains only one Cancel button.
- [x] Moving items between the LP root and chapters does not reload the complete builder panel.
- [x] The final certificate page works in course, session and group contexts.
- [x] Unassigned users/groups show a green plus; existing restrictions show a pencil.
- [x] Advanced access date selectors allow selecting date and time and persist both values.
- [x] The simple builder performance issue cannot be reproduced after warm-cache profiling.

PR #8735, opened 2026-07-15
Security: Skip unregistered roles in ResourceAclHelper instead of crashing

addow:
## Summary
- `ResourceAclHelper::isAllowed()` iterates every role in `$user->getRoles()` and calls `Acl::isAllowed($role, ...)` for each, but the Laminas ACL registry (built in `init()`) only ever registers a small fixed set of resource-permission roles (`ROLE_USER`, `ROLE_STUDENT`, `ROLE_CURRENT_COURSE_TEACHER`, ...).
- Any other role the user happens to carry - e.g. `ROLE_SESSION_MANAGER`, `ROLE_PREVIOUS_ADMIN` or `ROLE_ALLOWED_TO_SWITCH`, all present on an admin's token while using "Login as" - makes Laminas throw `Acl\Exception\InvalidArgumentException("Role '...' not found")`, crashing the whole request with a 500 instead of just being irrelevant to the resource-level permission check.
- This affects any `AbstractResource`-backed resource (documents, forums, etc.) whenever the current user's Symfony role set contains a role the ACL never registered - most commonly while impersonating another user via "Login as".
- Fix: skip roles the ACL hasn't registered via `Acl::hasRole()` instead of asking it about them.

## Test plan
- [x] Added `tests/CoreBundle/Helpers/ResourceAclHelperTest.php`, reproducing the crash with a mocked user carrying `ROLE_SESSION_MANAGER`/`ROLE_PREVIOUS_ADMIN`/`ROLE_ALLOWED_TO_SWITCH` - fails before the fix, passes after.
- [x] `vendor/bin/phpstan analyse` clean on the changed file (pre-existing unrelated findings on lines untouched by this change were verified present before this diff too).

### DIFF
--- a/src/CoreBundle/Helpers/ResourceAclHelper.php
+++ b/src/CoreBundle/Helpers/ResourceAclHelper.php
@@ -97,6 +97,19 @@ readonly class ResourceAclHelper
         $roles = $user instanceof UserInterface ? $user->getRoles() : [];
 
         foreach ($roles as $role) {
+            // init() only ever registers a small fixed set of resource-permission
+            // roles (ROLE_USER, ROLE_STUDENT, ROLE_CURRENT_COURSE_TEACHER, ...).
+            // $user->getRoles() returns the user's FULL Symfony role set, which
+            // for an admin using "Login as" also includes roles like
+            // ROLE_SESSION_MANAGER, ROLE_PREVIOUS_ADMIN or ROLE_ALLOWED_TO_SWITCH -
+            // none of which the ACL knows about. Laminas throws
+            // InvalidArgumentException for any role it hasn't registered, so
+            // without this guard a single unrelated role crashes the whole
+            // permission check instead of just being irrelevant to it.
+            if (!$acl->hasRole($role)) {
+                continue;
+            }
+
             if ($acl->isAllowed($role, $resourceLink->getId(), $askedMask)) {
                 return true;
             }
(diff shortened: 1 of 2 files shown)


AngelFQC (approved):
Approved. The diagnosis is right and the guard is the minimal fix: a role the ACL never registered cannot carry any rule (`allow()` would have thrown at registration time), so skipping it is equivalent to a deny-by-default lookup and cannot widen access.

Three non-blocking notes:

1. The test only proves "does not throw". With `$rights = []` nothing is granted, so replacing the `continue` with an early `return false` would keep it green. A second case where `$rights` grants `VIEW` to `ROLE_STUDENT` *after* the unregistered roles, asserting `true`, would pin the intent.

2. `init()` keeps the same exposure: `$acl->allow($right->getRole(), ...)` throws too if a `ResourceRight` row holds a role `init()` never registered. Not reachable today (rights are only created with `ROLE_TEACHER`/`ROLE_STUDENT`/`ROLE_USER`), but the symmetric guard is free.

3. The skip is silent now. A `logger->debug()` in that branch would keep the signal, since the root cause — the ACL not mirroring `security.yaml`'s `role_hierarchy` — is deliberately out of scope: expanding it would widen permissions on existing `resource_right` rows (e.g. `ROLE_HR` inheriting every `ROLE_TEACHER` right), which is a product decision, not a sync fix.

PR #8796, opened 2026-07-31
CI: Replace Behat with Playwright

ywarnier:


ywarnier:
Under review before merging into master

PR #8843, opened 2026-08-07
Internal: Restore MCP allowed roles value template

krista445:
## Summary

- add a focused fixture regression test so duplicate category keys cannot silently drop `mcp_allowed_roles`
- add an idempotent V300 repair migration for fresh 3.0 beta databases created during the broken-fixture window
- insert or refresh the template and link every matching settings row without changing administrator-selected values
- explicitly enable the SQLite extensions used by the migration regression tests

## Context

Upstream commit f4b1ed8526 fixed the duplicate `security` fixture key for future installs. The original MCP V300 migration already covers normal database upgrades, but a fresh install creates the current schema and loads fixtures without executing migrations. Fresh databases created between the MCP addition and the fixture fix could therefore retain a missing template after an in-place code update.

The forward repair belongs in V300 because MCP is a 3.0 feature. Its `down()` intentionally preserves repaired template and administrator data.

## Validation

- PHPUnit: 3 tests, 7 assertions
- PHP syntax checks
- ECS: no errors
- Psalm: no errors
- workflow YAML parse
- `git diff --check`

## Deployment note

The existing in-app update runner still hard-codes V210 migration paths. That stale version-policy restriction is tracked separately in #8845 rather than placing a 3.0 MCP repair in V210.

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -28,7 +28,7 @@ jobs:
               uses: shivammathur/setup-php@v2
               with:
                   php-version: ${{ matrix.php-versions }}
-                  extensions: mbstring, xml, ctype, iconv, intl, pdo, pdo_mysql, pdo_sqlite, sqlite3, dom, gd, json, soap, zip, bcmath
+                  extensions: mbstring, xml, ctype, iconv, intl, pdo, pdo_mysql, dom, gd, json, soap, zip, bcmath
                   ini-values: post_max_size=256M, max_execution_time=600, memory_limit=4096M, date.timezone=Europe/Paris
                   coverage: pcov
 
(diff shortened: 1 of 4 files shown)


ywarnier:
You PR includes the addition of the pdo_sqlite and sqlite3 extensions to the phpunit test suite requirements. We don't support sqlite at this point (as far as I know). Please update your PR to remove this change.

christianbeeznest:
I tested the repair logic from `Version20260807125041` against the current master using MariaDB/InnoDB.

I simulated the affected fresh-install state by removing the `mcp_allowed_roles` template and clearing its `value_template_id`. The migration logic correctly recreated one template, linked both the global and access-url settings, preserved `selected_value`, and remained idempotent when executed a second time.

After rollback, the database returned to its original state.

So the repair migration itself looks good to me.

I agree with Yannick's comment regarding SQLite, though. I would remove `pdo_sqlite`/`sqlite3` from the PHPUnit workflow and avoid making the migration test depend on SQLite. The existing MySQL test environment should be used instead.

Issue #8955, opened 2026-08-26
Complete Chamilo v3 Fill the blank question functionnalities with Chamilo 2.0.3 features


chamilo-uga:
In Chamilo 2.0.3 version in Fill the blank question some corrections and modifications has been done
- trim spaces of student answer
- deal with entities in wording and answer
- allow creation of menu choice with [un|two|three] syntax in Fill the blank wording.

<img width="699" height="699" alt="Image" src="https://github.com/user-attachments/assets/2a085803-86f2-4251-9edd-529a6f95604e" />

christianbeeznest:
This issue is being addressed by #8967.
I reviewed the PR and it covers the Fill in blanks changes requested here. There are still a few points to check/fix before merging, mainly translations, HTML entities handling and the current failing checks.
I will keep this issue open until the PR is updated, tested and merged.

christianbeeznest:
Completed in PR #8967.
Fill in the blanks now covers trimmed answers, dropdown syntax with |, multiple accepted answers with ||, and HTML entity handling, with additional tests included.
Closing this issue as resolved.

ywarnier:
@baelmyhu This one is important to test soon in the latest Git version to make sure it's good for v3 stable.

PR #8967, opened 2026-08-27
#8955 - Finalize Fill the blank question to fit 2.0.3 functionnalities

chamilo-uga:
Update Fill the blank Chamilo v3 to fit Chamilo 2.0.3's one

- trim student answer spaces
- allow syntax [Lima|Bogota|Paris] for creating menu
- allow syntax [1||one||un||uno] for acceptiong several answers
-

### DIFF
--- a/assets/locales/en_US.json
+++ b/assets/locales/en_US.json
@@ -3491,6 +3491,8 @@
     "Highest value": "Highest value",
     "Fill in blanks": "Fill in blanks",
     "Use the selected markers to define blanks. Example: The capital of Peru is [Lima].": "Use the selected markers to define blanks. Example: The capital of Peru is [Lima].",
+    "To turn a blank into a dropdown menu, separate the choices with a single |, the first one being the correct answer. Example: The capital of Peru is [Lima|Paris|Bogota].": "To turn a blank into a dropdown menu, separate the choices with a single {'|'}, the first one being the correct answer. Example: The capital of Peru is [Lima{'|'}Paris{'|'}Bogota].",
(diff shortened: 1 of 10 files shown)


ywarnier:
Hi @chamilo-uga 
2 questions: 
- the new language term in en_US.json already exists in translations/messages.pot and translations/messages.en_US.po? Otherwise, it has to be added there too (translations/ is the main translation repo, then assets/locales/ is just a filter of what we need in Vue)
- no database changes needed?

christianbeeznest:
Thanks, this looks like the right PR for #8955 and the | / || handling seems consistent with the migrated scorer.

Before merging, I think we should check a few points:

- The two new strings should also be added to translations/messages.pot and translations/messages.en_US.po, as Yannick mentioned.
- Could you please add/check a case with HTML entities in the answer? ExerciseRuntimeAnswerProcessor escapes the student value with htmlspecialchars(), while ExerciseAttemptScoringService decodes the expected answer before comparing it. A case like [AT&amp;T] with the student answering AT&T would be useful to verify.
- It would be good to have tests for trim, | dropdowns, || accepted answers and HTML entities.
- The current PR checks still show failures for ECS, PHPUnit and Playwright.

I don't see any database migration being necessary for this change.

christianbeeznest:
Reviewed and completed the remaining points in this PR.
I added answer normalization for whitespace and HTML entities, covered dropdown (|) and multiple accepted answers (||) cases, and added focused tests.
The missing translation entries were also completed.
Ready to merge.

PR #8977, opened 2026-08-28
Add MathJax formula support in TinyMCE

chamilo-uga:
This PR add the MathJax functionality in TinyMCE menu.
The icon to add a formula is Σ

You can enter math formula with the Latex Mathjax language.

image


chamilo-uga:
Found error

PR #8982, opened 2026-08-31
Add MathJax formula support in TinyMCE - #8977

chamilo-uga:
This PR add the MathJax functionality in TinyMCE menu.
The icon to add a formula is Σ

You can enter math formula with the Latex Mathjax language.

image


public/libs/editor/tinymce_plugins/latex/plugin.js
AngelFQC, line 1: This file is marked as binary 🤔

public/libs/mathjax/tex-svg.js
AngelFQC, line 1: mathjax ^4.1.2 is already a dependency here

Please copy it from node_modules at build time instead (Encore.copyFiles, next to the existing copy rules in webpack.config.js)

var/themes/chamilo/tiny-settings.js
```diff
@@ -535,16 +535,23 @@
...
+      "visualblocks wordcount latex",
     ].join(" "),
 
+    // "latex" isn't self-hosted under base_url/plugins/<name>/ like the others,
+    // so it needs an explicit entry — ensureExternalPluginsMap() below only
+    // fills in names that aren't already mapped.
+    external_plugins: {
+      latex: "/libs/editor/tinymce_plugins/latex/plugin.js",
+    },
+
```
AngelFQC, line 547: The latex plugin and its toolbar button are enabled unconditionally, but Chamilo already has a platform setting for exactly this: enabled_mathjax, declared in src/CoreBundle/Settings/EditorSettingsSchema.php

It is  already exposed to the frontend as `editor.enabled_mathjax`

src/CoreBundle/Resources/views/Layout/head.html.twig
```diff
@@ -167,6 +167,10 @@
     {# <script src="{{ asset('libs/readmore-js/readmore.js') }}"></script> #}
     {# <script src="{{ url('index') ~ 'build/libs/js-cookie/src/js.cookie.js' }}"></script> #}
 
+    {# Typesets stored LaTeX (span.math-latex from the "latex" TinyMCE plugin);
+       lazy-loads the actual MathJax library only when such content is on the page #}
+    <script src="{{ asset('libs/mathjax/mathjax-render.js') }}"></script>
```
AngelFQC, line 172: This template is reached by every page on the platform — base-layout.html.twig → no_layout.html.twig is what IndexController renders for the whole Vue SPA. That means mathjax-render.js installs a permanent MutationObserver on document.documentElement with subtree: true on every page, including SPA views that will never contain a formula. Two things would fix it: gate this <script> on the enabled_mathjax setting, and have the observer disconnect itself when a first scan finds no .math-latex node

public/libs/mathjax/mathjax-render.js
```diff
@@ -0,0 +1,131 @@
...
+        if (node.matches && node.matches(SELECTOR)) {
+          scan(node.parentNode || document)
+        } else if (node.querySelector && node.querySelector(SELECTOR)) {
+          scan(node)
+        }
+      })
+    })
+  })
+
+  observer.observe(document.documentElement, { childList: true, subtree: true })
```
AngelFQC, line 122: Observing `document.documentElement` with `subtree: true` for the lifetime of the page means the callback runs on every DOM insertion anywhere in the app
This is a Vue SPA, so that's a lot of them, on pages that mostly have no math at all

AngelFQC (changes requested):

AngelFQC:
Integrated in commit 4beeff13a4ce6cb468de5b35f237cd5d0c2cffc2

PR #8991, opened 2026-09-02
Drop three isset blocks in api_get_user_info_from_entity that assign a key to itself

darkdi:
`api_get_user_info_from_entity` has three `if (isset($result[X])) { $result[X] = $result[X]; }` blocks. Each reads the key and writes the same value back, so the guard and the body together do nothing at all.

The one worth a second look is `user_is_online_in_chat`, because the block directly above it is the same shape for `user_is_online` and does normalise the value to `1`/`0`. If both keys are supposed to reach the caller in the same form, then what is missing there is the cast, not the block, and I would rather you make that call than guess it. `profile_completed` and `extra` have no such neighbour, they look like leftovers.

I dropped all three, which keeps the returned array exactly as it is today. There are a few more of these in the legacy tree, I left them alone since they sit in unrelated files.

### DIFF
--- a/public/main/inc/lib/api.lib.php
+++ b/public/main/inc/lib/api.lib.php
@@ -1755,17 +1755,10 @@ function api_get_user_info_from_entity(
     if (isset($result['user_is_online'])) {
         $result['user_is_online'] = true == $result['user_is_online'] ? 1 : 0;
     }
-    if (isset($result['user_is_online_in_chat'])) {
-        $result['user_is_online_in_chat'] = $result['user_is_online_in_chat'];
-    }
 
     $result['password'] = '';
     if ($showPassword) {
         $result['password'] = $user->getPassword();
-    }
-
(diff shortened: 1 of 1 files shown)


AngelFQC (approved):
No cast is missing. `user_is_online_in_chat` is initialised to int `0` and only ever receives an int, so nothing was left to normalise — and the history confirms it: all three blocks were born dead in 81628ef26ea, the entity-style port of `_api_format_user` where the source `$user` mechanically became `$result`, and Psalm removed the one surviving `(int)` in a8affdffcee as double casting. Dropping all three is the right call. The real gap that port left behind is elsewhere — `picture_uri` and `profile_completed` are never set on this path, and the `$attributes` list right above is orphaned because the original `foreach` could not be ported to an entity — but that is a behaviour change, so I will open a separate issue for it. Thanks for the clean, scoped patch.

Issue #9084, opened 2026-09-17
Calcluated question : migrate legacy calculated question to new format. Follow issue #8194


chamilo-uga:
Issue #8194 upgarde calcluated questions.

We have now to migrate legacy calcluated questions format to new format.

Issue #8194, opened 2026-04-15
Upgrade of Calculated question


chamilo-uga:
Hi,
in our Chamilo, we did redo the Calculated question feature.
see https://chamilosimsu.wordpress.com/2025/02/11/questions-calculees-tirage-des-variables-ameliore/ (french)

We want to upgrade current Chamilo v2 Calculated question with our we've been using for many years in Chamilo v1. We've working on the Chamilo v2 version.

To do :
-> upgrade code from v1 to v2
-> create a tool that migrate every Calculated question to new Calculated question

<img width="1644" height="1800" alt="Image" src="https://github.com/user-attachments/assets/cccb5a81-23ce-44d1-9970-b36f9e3f83b2" />

chamilo-uga:
Difficilty of migration :
in v1, random interval is not saved in DB
As a result, I cannot get random interval to put it in the new DB data.
I'll put, for example an interval 0-1 for all, and users will have to check there Calculated answers question ?

chamilo-uga:
If you upgrade the code to tge new calculated answer question, here is the result :

<img width="765" height="421" alt="Image" src="https://github.com/user-attachments/assets/99115838-2fc6-4c62-8175-791e0002276d" />

PR #9085, opened 2026-09-17
#8194 V3 calculated formula : upgrade of Calculated question

chamilo-uga:
Fixes issue #8194
 upgrade of Calculated question

The migration process to migrate legacy Calculated question to this new question format well be in another PR that will fixes issue #9084


chamilo-uga:
I'll deal with unsuccessful checks.

chamilo-uga:
Will redo it including migration process

PR #9095, opened 2026-09-21
Sanitize segment names by replacing invalid characters

LudiscapeXApi:
Le nom de dossier Module 1: Finding Your Way Around Chamilo 3.0 contient : — invalide sur Windows. La correction se fait dans normalizeImportedDocumentRelativePath pour nettoyer les caractères interdits sur Windows dans chaque segment de chemin.


ywarnier:
## What the PR actually does

It's a single line added to normalizeImportedDocumentRelativePath() in src/CourseBundle/Component/CourseCopy/Moodle/Builder/MoodleImport.php (targets the 3.0 branch, not master):
```
               if (!$hasExtension && preg_match('~^(.+)-\d{3,}$~', $segment, $matches)) {
                   $segments[$index] = (string) $matches[1];
  +                $segments[$index] = (string) preg_replace('~[:<>"|?*]~', '-', (string) $segments[$index]);
               }
```

##  Does it change the title, or just the disk write?

It changes the title, not just disk I/O. I traced the caller: `normalizeImportedDocumentRelativePath()'s segments feed $ensureFolder()` (MoodleImport.php:1900-1949), where each segment becomes `$title = $seg;` passed straight into `DocumentManager::addDocument(...)` as the folder's title - the same value the user sees in the resource tree. It's also reused to build `$accum`, the logical path stored/used for lookup. So this is a rename of the imported folder, not a disk-encoding-only fix.

## Caveat on scope - the fix is narrower than its description claims

The new line only runs inside the existing `if (!$hasExtension && preg_match('~^(.+)-\d{3,}$~', ...))` branch - i.e., only for extension-less segments that already end in a numeric technical suffix like -23443 (Moodle's context-id suffix, which gets stripped first). It is not applied to every segment unconditionally, despite the PR body saying "nettoyer les caractères interdits... dans chaque segment de chemin":
- A folder/file name with a forbidden character but without a trailing -NNN suffix is left untouched.
- Any segment with a file extension ($hasExtension true) skips this whole block, so a document literally named My:File.pdf is not sanitized at all.
  
## Verdict

- Low regression risk: it's a simple, additive character-class replacement (`:<>"|?* → -`), no ReDoS, no unrelated code touched, additions=1/deletions=0.
- It does rename affected folders (title, not just storage), which is the intended/expected behavior for the reported bug.
- It's incomplete relative to its own description - only fixes the specific "technical numeric suffix" case, not general Windows-invalid characters in arbitrary segments. If we want full coverage, the preg_replace should be applied to every segment regardless of the -NNN/extension branch.

